PR #8, opened 2017-06-06
fix navigation extraction

effervescentia:


src/core/action-creator.ts
```diff
@@ -153,7 +153,7 @@ export default class Creator {
       dispatch(this.receiveQuery(Adapters.Search.extractQuery(results, this.linkMapper)));
       dispatch(this.receiveProducts(results.records.map(Adapters.Search.extractProduct)));
       // tslint:disable-next-line max-line-length
```
Vic-Dev: delete this you dummy :)

PR #11, opened 2017-06-14
BBD-572: Should trim query in searches

zydc2010:


src/core/action-creator.ts
```diff
@@ -121,8 +121,12 @@ export default class Creator {
 
   // request action creators
   updateSearch = (search: Actions.Search) =>
-    thunk<Actions.Search.UpdateSearch>(
-      Actions.UPDATE_SEARCH, search)
+    (dispatch: Dispatch<Actions.Search.UpdateSearch>) => {
+      const query = search.query ? search.query.trim() : '';
```
effervescentia: can be simplified to `const query = search.query && search.query.trim()`

test/unit/core/action-creator.ts
```diff
@@ -327,12 +327,35 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
-          const thunk = stub(utils, 'thunk');
+          const search: Actions.Search = { query: 'harry' };
+          const dispatch = spy();
+
+          actions.updateSearch(search)(dispatch);
+
+          expect(dispatch).to.be.calledWith({ type: Actions.UPDATE_SEARCH, ...<any>search });
+        });
+
+        it('should not create an UPDATE_SEARCH action when query string is empty', () => {
```
effervescentia: I would combine this test with the last one and just dispatch multiple "invalid" actions to show that none of them trigger a dispatch

src/core/action-creator.ts
jmbtutor: Updated

test/unit/core/action-creator.ts
jmbtutor: As previously discussed, we've decided against combining these because these test different expectations.

jmbtutor:
Requested changes have been implemented.

PR #12, opened 2017-06-15
BBD-570: Pagination can't navigate to last page on max result set

Guo-Haowei:
Server returns failure when the sum of pageSize and skip exceeds 10000.

### DIFF
--- a/test/unit/core/selectors.ts
+++ b/test/unit/core/selectors.ts
@@ -1,7 +1,30 @@
+import { MAX_RECORDS } from '../../../src/core/adapters/search';
 import Selectors from '../../../src/core/selectors';
 import suite from '../_suite';
 
 suite('selectors', ({ expect, stub }) => {
+
+  describe('searchRequest()', () => {
+    it('should decrease page size to prevent exceeding MAX_RECORDS', () => {
+      const remainingRecords = 2;
+      const originalPageSize = MAX_RECORDS - 1;
+      const originalSkip = MAX_RECORDS - remainingRecords;
+
+      stub(Selectors, 'fields');
+      stub(Selectors, 'query');
+      stub(Selectors, 'collection');
+      stub(Selectors, 'selectedRefinements');
+      stub(Selectors, 'sort');
+      stub(Selectors, 'requestSort');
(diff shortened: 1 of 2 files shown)


test/unit/core/selectors.ts
```diff
@@ -1,8 +1,34 @@
...
+    it('should decrease page size to prevent exceeding MAX_RECORDS', () => {
+      stub(Selectors, 'fields');
+      stub(Selectors, 'query');
+      stub(Selectors, 'collection');
+      stub(Selectors, 'selectedRefinements');
+      stub(Selectors, 'sort');
+      stub(Selectors, 'requestSort');
+
+      const originalPageSize = MAX_RECORDS - 1;
+      const state: any = {
```
effervescentia: I think you should be able to avoid passing a "real" state to this test if you stub `Selectors.pageSize()`,
good otherwise

effervescentia: As to why, it becomes a massive headache to change your data all over the place whenever you change the method contract (just ask searchandiser how much they like maintaining thousands of lines of JSON objects including escaping quotes)

jmbtutor: A headache indeed. We've stubbed the appropriate methods.

test/unit/core/selectors.ts
```diff
@@ -1,8 +1,29 @@
...
+    it('should decrease page size to prevent exceeding MAX_RECORDS', () => {
+      const originalPageSize = MAX_RECORDS - 1;
+
+      stub(Selectors, 'fields');
+      stub(Selectors, 'query');
+      stub(Selectors, 'collection');
+      stub(Selectors, 'selectedRefinements');
+      stub(Selectors, 'sort');
+      stub(Selectors, 'requestSort');
+      stub(Selectors, 'pageSize').returns(originalPageSize);
```
effervescentia: I would say having both of these return the same value, even if it conveniently showcases the behaviour, makes it confusing for future developers to read this as it implies that `pageSize` and `skip` somehow reference the same data. Equally it's unclear the the `pageSize` is `1` because we have removed `1` from `MAX_RECORDS`

PR #22, opened 2017-06-20
BBD-580: Add sort, collection, and details to URL beautifier

Vic-Dev:


Vic-Dev:
It's not ready!

src/core/action-creator.ts
```diff
@@ -196,8 +196,8 @@ export default class Creator {
         dispatch(this.receiveTemplate(Adapters.Search.extractTemplate(results.template))),
       );
 
-      Promise.all(updates)
-        .then(() => this.saveState());
+      return Promise.all(updates)
+        .then(() => this.saveState('search'));
```
effervescentia: I'd suggest we pull out `'search'` and the other routes everywhere they're used into a constant somewhere
(this includes places where it's used at the key in a map, like in the `mergeState` map)

src/core/adapters/configuration.ts
```diff
@@ -28,6 +29,9 @@ namespace Adapter {
       }
     });
 
+  export const extractArea = (config: FluxCapacitor.Configuration, defaultValue: string) =>
```
effervescentia: if you add a test for this guy it will probably push your coverage % over the one desired by bithound

src/core/action-creator.ts
Vic-Dev: ya good point

src/core/events.ts
```diff
@@ -73,3 +73,6 @@ export const TAG_ALIASING = 'tag:aliasing';
 
 // history events
 export const HISTORY_SAVE = 'history:save';
+
+// url events
+export const URL_UPDATED = 'url:updated';
```
effervescentia: are we still using this event?

effervescentia: I assume it's used somewhere in `@storefront/core`

PR #29, opened 2017-07-17
BBD-603: support additional SAYT options and language in request

effervescentia:


test/unit/core/selectors.ts
```diff
@@ -1,30 +1,275 @@
...
+  describe('area()', () => {
+    it('should select the global area from state', () => {
+      const area = 'myArea';
+      const state: any = { data: { area } };
+
+      expect(Selectors.area(state)).to.eq(area);
+    });
+  });
+
+  describe('area()', () => {
```
jmbtutor: `fields()`

test/unit/core/selectors.ts
```diff
@@ -1,30 +1,275 @@
...
+    it('should select the fields from state', () => {
+      const fields = ['a', 'b'];
+      const state: any = { data: { fields } };
+
+      expect(Selectors.fields(state)).to.eq(fields);
```
jmbtutor: Do we care that it's a reference to the same object?

test/unit/core/selectors.ts
```diff
@@ -37,6 +282,41 @@ suite('selectors', ({ expect, stub }) => {
...
+      expect(Selectors.skip(state, 12)).to.eq(36);
+    });
+  });
+
+  describe('products()', () => {
+    it('should select the products from the state', () => {
+      const products = ['a', 'b', 'd'];
+      const state: any = { data: { products } };
+
+      expect(Selectors.products(state)).to.eq(products);
```
jmbtutor: Same comment about reference to object, and same for `details` below.

src/core/selectors.ts
```diff
@@ -1,24 +1,62 @@
...
+      request.language = language;
+    }
+    if (sort) {
+      request.sort = <any>Selectors.requestSort(sort);
+    }
+
+    return <Request>{
+      ...config.search.defaults,
+      ...request,
+      ...config.search.overrides,
```
effervescentia: I want to make them mandatory at some point, because it makes editing in an IDE easier, but I'm fine with them being around messily at the moment

test/unit/core/adapters/configuration.ts
```diff
@@ -137,17 +141,83 @@ suite('SearchAdapter', ({ expect, stub }) => {
...
         allIds: [collectionDefault]
       });
     });
   });
+
+  describe('extractLanguage()', () => {
+    it('should return global language', () => {
+      const language = 'en';
+
+      expect(Adapter.extractLanguage(<any>{ language })).to.eql(language);
```
effervescentia: nice :P good catch, too much copy-pasta

test/unit/core/selectors.ts
effervescentia: I think unless we explicitly expect it to be a different object, it's always simpler (and faster) to check direct equality

PR #31, opened 2017-07-20
BBD-599 Range input controls for navigation

Guo-Haowei:


src/core/action-creator.ts
```diff
@@ -148,11 +148,28 @@ export function createActions(flux: FluxCapacitor) {
       updateSearch: (search: Actions.Payload.Search): Actions.Thunk<Actions.UpdateSearch> =>
         (dispatch) => {
           const query = search.query && search.query.trim();
-          if (query || query === null) {
-            dispatch(action(Actions.UPDATE_SEARCH, { ...<any>search, query }, metadata));
+          if (!('query' in search) || (query || query === null)) {
+            const payload: any = 'query' in search ? { ...search, query } : search;
```
effervescentia: see if making the type here `Actions.Payload.Search` instead of `any` works

src/core/reducers/data/navigations.ts
```diff
@@ -54,7 +48,7 @@ export const updateSearch = (state: State, payload: Actions.Payload.Search) => {
     } else {
       return addRefinement({ ...state, byId }, <Actions.Payload.Navigation.AddRefinement>payload);
     }
-  } else {
+  } else if (payload.clear) {
```
effervescentia: it looks like we're missing one of the test cases here as we only have partial coverage of this line

src/core/reducers/data/query.ts
```diff
@@ -18,8 +18,8 @@ export default function updateQuery(state: State = DEFAULTS, action: Action): St
   }
 }
 
-export const updateOriginal = (state: State, { query: original }: Actions.Payload.Search) =>
-  ({ ...state, original });
+export const updateOriginal = (state: State, payload: Actions.Payload.Search) =>
+  'query' in payload ? ({ ...state, original: payload.query }) : state;
```
effervescentia: partial coverage

PR #35, opened 2017-07-24
Add recommendations API results to SAYT

effervescentia:


src/core/configuration.ts
```diff
@@ -87,81 +172,51 @@ interface Configuration {
...
-    errorHandler?: (error: Error) => void;
+    idField: string;
     /**
-     * add SkipCache header to search requests
-     */
-    skipCache?: boolean;
-    /**
-     * add SkipSemantish header to search requests
+     * type of product recommendations to request
+     * can be 'popular', 'trending' or 'recent'
```
Vic-Dev: seems unnecessary to state this

src/core/reducers/session.ts
```diff
@@ -4,20 +4,29 @@ import Store from '../store';
...
-  return state;
+      return state;
+    }
+  }
 }
 
+export const updateLocation = (state: State, location: Store.Location) =>
+  ({ ...state, location });
+
 export const updateRecallId = (state: State, { recallId }: Actions.Metadata) =>
```
Vic-Dev: why cant one function?

Vic-Dev: instead 3

Vic-Dev: of

src/core/sagas/autocomplete.ts
```diff
@@ -2,21 +2,66 @@ import * as effects from 'redux-saga/effects';
...
+      const field = Selectors.autocompleteCategoryField(state);
+      const location = Selectors.location(state);
+      const suggestionsRequest = effects.call(
         [flux.clients.sayt, flux.clients.sayt.autocomplete],
         query,
         Selectors.autocompleteSuggestionsRequest(flux.config)
       );
-      const suggestions = Adapter.extractSuggestions(res, field);
+      const config = flux.config.autocomplete.recommendations;
+      const suggestionMode = Configuration.RECOMMENDATION_MODES[config.suggestionMode || 'popular'];
```
Vic-Dev: should probably add a comment that we're defaulting to popular bc not really clear why we're doing this

src/core/sagas/recommendations.ts
```diff
@@ -0,0 +1,52 @@
...
+import Store from '../store';
+import { fetch } from '../utils';
+
+export namespace Tasks {
+  // tslint:disable-next-line max-line-length
+  export function* fetchProducts(flux: FluxCapacitor, action: Actions.FetchRecommendationsProducts) {
+    try {
+      const state = yield effects.select();
+      const config = flux.config.recommendations;
+      const mode = Configuration.RECOMMENDATION_MODES[config.mode || 'popular'];
```
Vic-Dev: Again should comment/explain the or

src/core/sagas/autocomplete.ts
```diff
@@ -2,21 +2,66 @@ import * as effects from 'redux-saga/effects';
...
+      // tslint:disable-next-line max-line-length
+      const trendingUrl = `https://${flux.config.customerId}.groupbycloud.com/wisdom/v2/public/recommendations/searches/_get${suggestionMode}`;
```
jmbtutor: Can we pull this out into another module? It feels wrong having raw a URL here.

src/core/sagas/autocomplete.ts
```diff
@@ -2,21 +2,66 @@ import * as effects from 'redux-saga/effects';
...
 
 export namespace Tasks {
+  // tslint:disable-next-line max-line-length
   export function* fetchSuggestions(flux: FluxCapacitor, { payload: query }: Actions.FetchAutocompleteSuggestions) {
     try {
-      const field = yield effects.select(Selectors.autocompleteCategoryField);
-      const res = yield effects.call(
+      const state = yield effects.select();
+      const field = Selectors.autocompleteCategoryField(state);
+      const location = Selectors.location(state);
```
jmbtutor: It won't actually conflict due to scope, but should we be shadowing the global `location` object given that this is eventually going to be run in a browser?

src/core/sagas/autocomplete.ts
```diff
@@ -2,21 +2,66 @@ import * as effects from 'redux-saga/effects';
...
+      const trendingUrl = `https://${flux.config.customerId}.groupbycloud.com/wisdom/v2/public/recommendations/searches/_get${suggestionMode}`;
+      const trendingBody: any = {
+        size: config.suggestionCount,
+        matchPartial: {
+          and: [{
+            search: { query }
+          }]
+        }
+      };
+      if (location && config.location) {
```
jmbtutor: Does `config.location` have any significance here?

src/core/sagas/recommendations.ts
```diff
@@ -0,0 +1,52 @@
...
+      // tslint:disable-next-line max-line-length
+      const recommendationsUrl = `https://${flux.config.customerId}.groupbycloud.com/wisdom/v2/public/recommendations/products/_get${mode}`;
```
jmbtutor: Same comment about URL.

src/core/store/index.ts
```diff
@@ -281,6 +293,11 @@ namespace Store {
       selected?: string;
     }
   }
+
+  export interface Location {
```
jmbtutor: Perhaps rename to `Geolocation` to avoid ambiguity between browser `location` or some other location.

src/core/sagas/autocomplete.ts
effervescentia: I think this is fine, we should only ever use `location` and the rest of the globals through `WINDOW()`

src/core/sagas/autocomplete.ts
effervescentia: agreed

src/core/sagas/autocomplete.ts
effervescentia: determines whether or not the location should be requested in the first place by the autocomplete service, but I believe it's redundant here, yes

src/core/configuration.ts
effervescentia: forgot that it would generate the type info 👍

src/core/reducers/session.ts
effervescentia: I will make a ticket to refactor the reducers to be simpler

effervescentia: BBD-620

src/core/sagas/autocomplete.ts
```diff
@@ -17,12 +19,13 @@ export namespace Tasks {
       const suggestionsRequest = effects.call(
         [flux.clients.sayt, flux.clients.sayt.autocomplete],
         query,
-        Selectors.autocompleteSuggestionsRequest(flux.config)
+        Requests.autocompleteSuggestions(flux.config)
       );
       const config = flux.config.autocomplete.recommendations;
+      // fall back to default mode "popular" if not provided
```
jmbtutor: This mentions what it's doing, but not why. I think @Vic-Dev meant explaining why we fall back to "popular" as opposed to the other modes.

src/core/requests.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    return <Request>{
+      ...config.search.defaults,
+      ...request,
+      ...config.search.overrides,
+    };
+  };
+
+  export const autocompleteSuggestions = (config: AppConfig): QueryTimeAutocompleteConfig => {
+    return {
```
jmbtutor: Expression syntax? For `autocompleteProducts` below as well.

src/core/sagas/recommendations.ts
```diff
@@ -13,9 +14,10 @@ export namespace Tasks {
     try {
       const state = yield effects.select();
       const config = flux.config.recommendations;
+      // fall back to default mode "popular" if not provided
```
jmbtutor: Same comment from before.

jmbtutor:
Looks good to me.

PR #37, opened 2017-07-25
BBD-599 Range input controls for navigation

Guo-Haowei:


src/core/action-creator.ts
```diff
@@ -50,14 +50,14 @@ export function createActions(flux: FluxCapacitor) {
 
       addRefinement: (field: string, valueOrLow: any, high: any = null) => {
         if (typeof high === 'number') {
-          actions.updateSearch({
+          return actions.updateSearch({
```
effervescentia: we should add tests to show that we expect this to return so we don't have this problem again

test/unit/core/action-creator.ts
```diff
@@ -139,24 +139,29 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
-        const updateSearch = actions.updateSearch = spy();
         const navigationId = 'book';
         const value = 'a';
 
-        actions.addRefinement(navigationId, value);
+        const result = actions.addRefinement(navigationId, value);
 
-        expect(updateSearch).to.be.calledWith({ navigationId, value });
+        expect(result.type).to.be.eq('UPDATE_SEARCH');
+        expect(result.payload).to.be.eql({ navigationId, value });
```
effervescentia: instead, mock the `updateSearch` action and return a mock that you can compare here, we don't want this to break when the functionality of `updateSearch` changes

test/unit/core/action-creator.ts
```diff
@@ -139,24 +139,29 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
       });
 
       it('should update a range refinement when three parameters are provided', () => {
-        const updateSearch = actions.updateSearch = spy();
         const navigationId = 'book';
         const low = 0;
         const high = 1;
 
-        actions.addRefinement(navigationId, low, high);
+        const result = actions.addRefinement(navigationId, low, high);
```
effervescentia: same here

PR #40, opened 2017-07-26
BBD-631: support switchRefinement action

effervescentia:


src/core/reducers/data/navigations.ts
```diff
@@ -26,11 +26,25 @@ export default function updateNavigations(state: State = DEFAULTS, action: Actio
 }
 
 export const updateSearch = (state: State, payload: Actions.Payload.Search) => {
-  if (payload.clear) {
-    const byId = state.allIds.reduce((navs, nav) =>
-      Object.assign(navs, { [nav]: { ...state.byId[nav], selected: [] } }), {});
+  if (payload.clear in state.byId) {
+    const navigationId = <string>payload.clear;
```
effervescentia: `payload.clear` will only be handled here if it's an existing navigation id, and thus is a field name
We would suffer here if the field name was `true` or `false`

src/core/reducers/data/navigations.ts
```diff
@@ -26,11 +26,25 @@ export default function updateNavigations(state: State = DEFAULTS, action: Actio
...
 
-    state = { ...state, byId };
+    state = {
+      ...state,
+      byId: {
+        ...state.byId,
+        [navigationId]: {
```
effervescentia: should always be a valid field name here

src/core/reducers/data/navigations.ts
```diff
@@ -26,11 +26,25 @@ export default function updateNavigations(state: State = DEFAULTS, action: Actio
...
+          ...state.byId[navigationId],
+          selected: []
+        }
+      }
+    };
+  } else if (payload.clear) {
+    state = {
+      ...state,
+      byId: state.allIds.reduce((navs, nav) =>
```
effervescentia: can't use `map()` as we're creating an object

src/core/reducers/data/navigations.ts
```diff
@@ -26,11 +26,25 @@ export default function updateNavigations(state: State = DEFAULTS, action: Actio
...
+        Object.assign(navs, { [nav]: { ...state.byId[nav], selected: [] } }), {})
```
jmbtutor: Although I realize this was existing code, and this suggestion may be premature optimization, but I would do a direct assignment since it would be much more performant than doing an `Object.assign` for every nav.
```js
(navs, nav) => {
  navs[nav] = { ... state.byId[nav], selected: [] } };
  return navs;
}
```

Alternatively, as a one-liner:

```js
(navs, nav) => ((navs[nav] = { ... state.byId[nav], selected: [] } }) || nav)
```

I think the first form is clearer.

PR #42, opened 2017-07-27
BBD-635: skip recommendations if suggestion count is 0

effervescentia:


test/unit/core/sagas/autocomplete.ts
```diff
@@ -80,6 +80,36 @@ suite('autocomplete saga', ({ expect, spy, stub }) => {
...
+        stub(Requests, 'autocompleteSuggestions').returns(request);
+        stub(Selectors, 'location');
+        stub(Selectors, 'autocompleteCategoryField');
+        stub(Adapter, 'extractSuggestions').returns(suggestions);
+        // tslint:disable-next-line max-line-length
+
+        const task = Tasks.fetchSuggestions(flux, <any>{ payload: query });
+
+        task.next();
+        expect(task.next(state).value).to.eql(effects.all([effects.call([sayt, autocomplete], query, request)]));
```
zydc2010: I think `// tslint:disable-next-line max-line-length` should be on line 107 instead of line 102.

PR #49, opened 2017-08-11
feat(BBD-623): collapsible nav component

Vic-Dev:


src/core/selectors.ts
```diff
@@ -133,6 +138,12 @@ namespace Selectors {
 
   export const recommendationsProducts = (state: Store.State) =>
     state.data.present.recommendations.products;
+
+  export const uiTagName = (state: Store.State, tagName: string) =>
```
effervescentia: I think a more appropriate name for this would be `tagStates()` or `uiTagStates()`, because that's the name of what you're actually selecting

src/core/selectors.ts
```diff
@@ -133,6 +138,12 @@ namespace Selectors {
...
+    state.ui[tagName];
+
+  export const tagId = (state: Store.State, tagName: string, id: string) =>
```
effervescentia: `tagState()` or `uiTagState()`

test/unit/core/observer.ts
```diff
@@ -489,4 +489,29 @@ suite('Observer', ({ expect, spy, stub }) => {
...
+  describe('ui', () => {
+    let emit;
+    let observers;
+
+    beforeEach(() => {
+      emit = spy();
+      observers = Observer.create(<any>{ emit });
+    });
+
+    it('should emit UI_ISACTIVE_UPDATED event', () => {
```
effervescentia: event name

test/unit/core/reducers/ui.ts
```diff
@@ -37,6 +37,19 @@ suite('ui', ({ expect }) => {
...
+      const reducer = ui(<any>{ a: 'b' }, { type: Actions.UPDATE_SEARCH, payload: { query: 'eyy' } });
+
+      expect(reducer).to.eql({});
+    });
+
+    it('should return state on UPDATE_SEARCH when no query in payload', () => {
+      const newState = <any>{ a: 'b' };
+      const reducer = ui(newState, { type: Actions.UPDATE_SEARCH, payload: {} });
+
+      expect(reducer).to.eql(newState);
```
effervescentia: should be `eq()`?

PR #51, opened 2017-08-15
feat(BBD-636): StoreFront should read the base path from the <base> tag

zydc2010:
https://github.com/groupby/storefront-core/pull/56


zydc2010:
No longer need to save baseUrl in state

PR #56, opened 2017-08-17
fix(BBD-638): Query returns all the results after clicking on '×' button

Guo-Haowei:


src/core/action-creator.ts
```diff
@@ -47,10 +47,16 @@ export function createActions(flux: FluxCapacitor) {
...
-                msg: 'search term is empty'
-              }
+              payload: [
+                {
+                  func: ({ query }) => !('query' in search) || !!query || query === null,
+                  msg: 'search term is empty'
+                },
+                {
+                  func: ({ query }, state) => query !== Selectors.query(state),
+                  msg: 'search term is same as previous one'
```
effervescentia: can we change this to `'search term is not different'`?

src/core/action-creator.ts
```diff
@@ -47,10 +47,16 @@ export function createActions(flux: FluxCapacitor) {
...
             validator: {
-              payload: {
-                func: ({ query }) => !('query' in search) || !!query || query === null,
-                msg: 'search term is empty'
-              }
+              payload: [
+                {
+                  func: ({ query }) => !('query' in search) || !!query || query === null,
+                  msg: 'search term is empty'
+                },
```
effervescentia: typically we try to keep the brackets on the same line for closing one object and opening the next
ex: `}, {`

PR #64, opened 2017-09-08
BBD 711: Update SAYT categories to have the right format

zfcb:
Also requires related PR for storefront-sayt to be merged:

https://github.com/groupby/storefront-sayt/pull/30

### DIFF
--- a/src/core/adapters/autocomplete.ts
+++ b/src/core/adapters/autocomplete.ts
@@ -23,9 +23,13 @@ namespace Adapter {
     const searchTerms = result.searchTerms || [];
     const navigations = result.navigations || [];
     const hasCategory = category && searchTerms[0] && Adapter.termsMatch(searchTerms[0].value, query);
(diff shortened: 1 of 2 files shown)


src/core/adapters/autocomplete.ts
```diff
@@ -23,9 +23,13 @@ namespace Adapter {
     const searchTerms = result.searchTerms || [];
     const navigations = result.navigations || [];
     const hasCategory = category && searchTerms[0] && Adapter.termsMatch(searchTerms[0].value, query);
+    // tslint:disable-next-line max-line-length
+    const categoryValues = hasCategory ? [{ matchAll: true }, ...Adapter.extractCategoryValues(searchTerms[0], category)] : [];
+    if (hasCategory) {
```
effervescentia: I don't think we test that `searchTerms` gets shifted

Filip-F: New test added to check for this.

PR #66, opened 2017-09-13
fix(BBD-712): Selecting multiple query + refinements in SAYT fails to update the page

zfcb:


src/core/action-creator.ts
```diff
@@ -51,7 +51,21 @@ export function createActions(flux: FluxCapacitor) {
...
+                  let currentRefinements = Selectors.selectedRefinements(state);
+                  const searchHasRefinement = Boolean(value && navigationId);
+                  // query is not the same
+                  return query !== Selectors.query(state) ||
+                  // xor for either there exists refinements or the current search has refinement
+                  currentRefinements.length > 0 !== searchHasRefinement ||
+                  // current search has refinement and does not match any of the previous refinements
+                  searchHasRefinement &&
+                    // try to find any refinement not matching the refinement in the query
+                    Selectors.selectedRefinements(state).find(
```
effervescentia: do we need to re-select `Selector.selectedRefinements(state)` when we have `currentRefinements`?

src/core/action-creator.ts
```diff
@@ -51,7 +51,21 @@ export function createActions(flux: FluxCapacitor) {
                 func: ({ query }) => !('query' in search) || !!query || query === null,
                 msg: 'search term is empty'
               }, {
-                func: ({ query }, state) => query !== Selectors.query(state) || query === null,
+                  // currently assume that search will only ever have up to one refinement
+                  func: ({ query, value, navigationId }, state) => {
+                  let currentRefinements = Selectors.selectedRefinements(state);
```
effervescentia: does the value of `currentRefinements` change? could be a `const`

src/core/action-creator.ts
```diff
@@ -51,7 +51,21 @@ export function createActions(flux: FluxCapacitor) {
...
+                  const searchHasRefinement = Boolean(value && navigationId);
```
effervescentia: don't need to coerce this to a boolean, we just need it to be truthy

test/unit/core/action-creator.ts
```diff
@@ -147,6 +147,7 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
         const search: any = { a: 'b' };
         const state = { a: 'b' };
         stub(Selectors, 'query').withArgs(state).returns(query);
+        stub(Selectors, 'selectedRefinements').withArgs(state).returns([]);
```
effervescentia: don't need to specify `withArgs()` here, we should only use this when we care about returning multiple distinct results based on the input, but as `selectedRefinements()` is only called once.
Instead it would be easier to include a test that explicitly expects `selectedRefinements` to be called with the `state`, that way we don't need to include it in all the other tests

```js
const selectedRefinements = stub(Selectors, 'selectedRefinements').returns([]);

// ...

expect(selectedRefinements).to.be.calledWith(state)`
```

src/core/action-creator.ts
```diff
@@ -51,7 +51,21 @@ export function createActions(flux: FluxCapacitor) {
                 func: ({ query }) => !('query' in search) || !!query || query === null,
                 msg: 'search term is empty'
               }, {
-                func: ({ query }, state) => query !== Selectors.query(state) || query === null,
+                  // currently assume that search will only ever have up to one refinement
+                  func: ({ query, value, navigationId }, state) => {
```
effervescentia: there are actually a few more options for the parameters to this function, `low`, `high`, `range` and `refinementId`

src/core/action-creator.ts
```diff
@@ -51,7 +51,21 @@ export function createActions(flux: FluxCapacitor) {
...
+                  const searchHasRefinement = Boolean(value && navigationId);
+                  // query is not the same
+                  return query !== Selectors.query(state) ||
+                  // xor for either there exists refinements or the current search has refinement
+                  currentRefinements.length > 0 !== searchHasRefinement ||
+                  // current search has refinement and does not match any of the previous refinements
+                  searchHasRefinement &&
+                    // try to find any refinement not matching the refinement in the query
+                    Selectors.selectedRefinements(state).find(
+                      (refinement) => (refinement.value !== value ||
```
effervescentia: the logic here needs to include cases for selecting a refinement by id, or a range refinement

zfcb:
@effervescentia 

Requested changes have been made!

src/core/action-creator.ts
```diff
@@ -51,7 +51,33 @@ export function createActions(flux: FluxCapacitor) {
                 func: ({ query }) => !('query' in search) || !!query || query === null,
                 msg: 'search term is empty'
               }, {
-                func: ({ query }, state) => query !== Selectors.query(state) || query === null,
+                  // currently assume that search will only ever have up to one refinement
+                  func: ({ query, value, navigationId, range, low, high, index }, state) => {
```
effervescentia: could be:

```js
import SearchAdapter from './adapters/search';

function refinementsChanged(clear, navigationId, index, value, low, high, state) {
  const currentRefinements = Selectors.selectedRefinements(state);
  const hasIndex = index >= 0;
  if (clear) {
    const clearedNavigation = Selectors.navigation(state, clear);
    if (clearedNavigation) {
      if (clear === navigationId) {
        if (clearedNavigation.refinements.length !== 1) {
          return true;
        } else if (hasIndex) {
          return isRefinementDeselected(state, navigationId, index);
        } else {
          return SearchAdapter.refinementsMatch({ low, high, value }, clearedNavigation.refinements[0], clearedNavigation.range || 'Value');
        }
      } else {
        return clearedNavigation.refinements.length !== 0;
      }
    } else if (currentRefinements.length === 1) {
      if (hasIndex) {
        return isRefinementDeselected(state, navigationId, index);
      } else {
        return SearchAdapter.refinementsMatch({ low, high, value }, currentRefinements[0]);
      }
    } else {
      return currentRefinements.length !== 0;
    }
  } else if (hasIndex) {
    return isRefinementDeselected(state, navigationId, index);
  } else {
    return !currentRefinements.find((refinement) => {
      return refinement.navigationName === navigationId && SearchAdapter.refinementsMatch({ low, high, value }, refinement);
    });
  }
}

func: ({ query, value, navigationId, low, high, index, clear }, state) => {
  const queryHasChanged = query !== Selectors.query(state);
  const refinementsHaveChanged = (navigationId || clear) && refinementsChanged(clear, navigationId, index, value, low, high, state);

  return queryHasChanged || refinementsHaveChanged;
}
```

effervescentia: as you'll notice above, we still need to handle the case where we pass `clear` to indicate that we intend to clear all the refinements first, `clear` can be a boolean or the field of a `navigation`

effervescentia: I plan to simplify the above action creator to turn it into a bulk action creator, I'll sit with you in the afternoon to go over updating it and the validator

test/unit/core/action-creator.ts
```diff
@@ -155,20 +156,203 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
+        const query = 'a';
+        const search: any = { query: 'a' };
+        const refinement = { navigationId: 'aaa', index: 0 };
+        const state = {};
+        stub(Selectors, 'query').returns(query);
+        stub(Selectors, 'selectedRefinements').returns(
+          [{ navigationName: 'aa', value: 'bb', range: true, low: 1, high: 2}]);
+        stub(Selectors, 'refinementCrumb').returns(
+          { navigationName: 'aa', value: 'bb', range: true, low: 1, high: 3});
+        expectAction(() => actions.updateSearch(search), Actions.UPDATE_SEARCH, search,
```
effervescentia: should separate the expectation and the setup with a newline

test/unit/core/action-creator.ts
```diff
@@ -155,20 +156,203 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
+          const query = 'a';
+          const search: any = { query: 'a' };
+          const refinement = { navigationId: 'aa', index: 0 };
+          const state = {};
+          stub(Selectors, 'query').returns(query);
+          stub(Selectors, 'selectedRefinements').returns(
+            [{ navigationName: 'aa', value: 'bb', range: true, low: 1, high: 2}]);
+          stub(Selectors, 'refinementCrumb').returns(
+            { navigationName: 'aa', value: 'bb', range: true, low: 1, high: 2});
+          expectAction(() => actions.updateSearch(search), Actions.UPDATE_SEARCH, search,
```
effervescentia: newline

test/unit/core/action-creator.ts
```diff
@@ -155,20 +156,203 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
+        const query = 'a';
+        const search: any = { query: 'a' };
+        const refinement = { navigationId: 'aa', range: true, low: 1, high: 2 };
+        const state = {};
+        stub(Selectors, 'query').returns(query);
+        stub(Selectors, 'selectedRefinements').returns(
+          [{ navigationName: 'aa', value: 'bb', range: true, low: 1, high: 2}]);
+        stub(Selectors, 'refinementCrumb').returns(
+          [{ navigationName: 'aa', value: 'bb', range: true, low: 1, high: 2}]);
+        expectAction(() => actions.updateSearch(search), Actions.UPDATE_SEARCH, search,
```
effervescentia: newline

test/unit/core/action-creator.ts
```diff
@@ -155,20 +156,203 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
+      it('should return an action with validation that evaluates to ' +
+      'true if low value for the range is different', () => {
+        const query = 'a';
+        const search: any = { query: 'a' };
+        const refinement = { navigationId: 'aa', range: true, low: 3, high: 2 };
+        const state = {};
+        stub(Selectors, 'query').returns(query);
+        stub(Selectors, 'selectedRefinements').returns(
+          [{ navigationName: 'aa', value: 'bb', range: true, low: 1, high: 2}]);
+        expectAction(() => actions.updateSearch(search), Actions.UPDATE_SEARCH, search,
```
effervescentia: newline

test/unit/core/action-creator.ts
```diff
@@ -155,20 +156,203 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
+        stub(Selectors, 'selectedRefinements').returns([]);
 
         expectAction(() => actions.updateSearch(search), Actions.UPDATE_SEARCH, search,
           (meta) => expect(meta.validator.payload[1].func({ query: 'boot' }, state)).to.be.true);
       });
 
-      it('should return an action with validation if search term is null', () => {
+      it('should return an action with validation that evaluates to ' +
+        'false if search term is null and refinements are same', () => {
         const query = null;
```
effervescentia: we should try to pull as many of these `const`s as possible into the containing `describe()` to simplify each `it()`s setup

effervescentia: I'm thinking `state`, `query`, `search`, `refinement`

test/unit/core/action-creator.ts
```diff
@@ -155,20 +156,203 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
+      it('should return an action with validation that evaluates to ' +
+      'true if high value for the range is different', () => {
+        const query = 'a';
+        const search: any = { query: 'a' };
+        const refinement = { navigationId: 'aa', range: true, low: 1, high: 222 };
+        const state = {};
+        stub(Selectors, 'query').returns(query);
+        stub(Selectors, 'selectedRefinements').returns(
+          [{ navigationName: 'aa', value: 'bb', range: true, low: 1, high: 2}]);
+        expectAction(() => actions.updateSearch(search), Actions.UPDATE_SEARCH, search,
```
effervescentia: newline

test/unit/core/action-creator.ts
```diff
@@ -155,20 +156,203 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
+      it('should return an action with validation that evaluates to ' +
+      'true if both low and high values for the range are different', () => {
+        const query = 'a';
+        const search: any = { query: 'a' };
+        const refinement = { navigationId: 'aa', range: true, low: 3, high: 22 };
+        const state = {};
+        stub(Selectors, 'query').returns(query);
+        stub(Selectors, 'selectedRefinements').returns(
+          [{ navigationName: 'aa', value: 'bb', range: true, low: 1, high: 2}]);
+        expectAction(() => actions.updateSearch(search), Actions.UPDATE_SEARCH, search,
```
effervescentia: newline

test/unit/core/action-creator.ts
```diff
@@ -155,20 +156,203 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
+      it('should return an action with validation that evaluates to ' +
+      'true if both low and high values for the range are same but navigationId is different', () => {
+        const query = 'a';
+        const search: any = { query: 'a' };
+        const refinement = { navigationId: 'ab', range: true, low: 1, high: 2 };
+        const state = {};
+        stub(Selectors, 'query').returns(query);
+        stub(Selectors, 'selectedRefinements').returns(
+          [{ navigationName: 'aa', value: 'bb', range: true, low: 1, high: 2}]);
+        expectAction(() => actions.updateSearch(search), Actions.UPDATE_SEARCH, search,
```
effervescentia: newline

test/unit/core/action-creator.ts
```diff
@@ -155,20 +156,203 @@ suite('ActionCreator', ({ expect, spy, stub }) => {
...
+
+      it('should call Selectors.selectedRefinements and Selectors.query with state', () => {
+        const query = 'a';
+        const search = { query: 'a' };
+        const refinement = { navigationId: 'aa', range: true, low: 1, high: 2 };
+        const state = {};
+        const selectedRefinements = stub(Selectors, 'selectedRefinements').returns([]);
+        const selectorsQuery = stub(Selectors, 'query').returns(query);
+        actions.updateSearch(search);
+        expectAction(() => actions.updateSearch(search), Actions.UPDATE_SEARCH, search,
```
effervescentia: newline

PR #68, opened 2017-09-19
fix(BBD 723): refactor UPDATE_SEARCH action to be a bulk action creator

Filip-F:
Since 712 got cancelled and the fixes happened here this also fixes BBD 712: selecting multiple query + refinements in SAYT fails to update the page.

closes #66

### DIFF
--- a/src/core/action-creator.ts
+++ b/src/core/action-creator.ts
@@ -42,29 +42,30 @@ export function createActions(flux: FluxCapacitor) {
         action(Actions.FETCH_RECOMMENDATIONS_PRODUCTS, null, metadata),
 
       // request action creators
-      updateSearch: (search: Actions.Payload.Search): any => {
-        const searchActions: any[] = [actions.resetPage()];
+      updateSearch: (search: Actions.Payload.Search): Actions.UpdateSearch => {
+        const searchActions: any = [actions.resetPage()];
 
         if ('query' in search) {
-          searchActions.push(actions.updateQuery(search.query));
+          searchActions.push(...actions.updateQuery(search.query));
         }
         if ('clear' in search) {
-          searchActions.push(actions.resetRefinements(search.clear));
+          searchActions.push(...actions.resetRefinements(search.clear));
         }
         if ('navigationId' in search) {
           if ('index' in search) {
-            searchActions.push(actions.selectRefinement(search.navigationId, search.index));
+            searchActions.push(...actions.selectRefinement(search.navigationId, search.index));
           } else if (search.range) {
-            searchActions.push(actions.addRefinement(search.navigationId, search.low, search.high));
+            searchActions.push(...actions.addRefinement(search.navigationId, search.low, search.high));
           } else if ('value' in search) {
-            searchActions.push(actions.addRefinement(search.navigationId, search.value));
+            searchActions.push(...actions.addRefinement(search.navigationId, search.value));
           }
         }
 
         return searchActions;
       },
 
-      updateQuery: (query: string): Actions.UpdateQuery =>
+      updateQuery: (query: string): Actions.ResetPageAndUpdateQuery => [
+        actions.resetPage(),
         action(Actions.UPDATE_QUERY, query && query.trim(), {
           ...metadata,
           validator: {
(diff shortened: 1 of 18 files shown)


src/core/action-creator.ts
```diff
@@ -41,38 +42,140 @@ export function createActions(flux: FluxCapacitor) {
...
+          }
+        }
+
+        return searchActions;
+      },
 
-      addRefinement: (field: string, valueOrLow: any, high: any = null) =>
-        actions.updateSearch(refinementPayload(field, valueOrLow, high)),
+      checkAndResetRefinements: ({ low, high, value, navigationId, range, clear }: Actions.Payload.Search):
+        Actions.CheckAndResetRefinements => {
```
effervescentia: rather than make this an action creator, why not make it a discriminator? like `shouldResetRefinements()` which just returns a `boolean`. That way you can just add it to the `if` check for adding `resetRefinements()` to the `updateSearch()` batch action

src/core/action-creator.ts
```diff
@@ -41,38 +42,140 @@ export function createActions(flux: FluxCapacitor) {
...
-        actions.updateSearch({ query, navigationId, index, clear: true }),
+      search: (query: string = Selectors.query(flux.store.getState())): Actions.Search => <any>[
+        actions.resetPage(),
+        ...actions.resetRefinements(true),
+        ...actions.updateQuery(query)
+      ],
+      // tslint:disable-next-line max-line-length
+      resetRecall: (query: string = null, { field, index }: { field: string, index: number } = <any>{}): Actions.ResetRecall => {
+        const resetActions: any = [
+          actions.resetPage(),
```
effervescentia: this initial value is equivalent to spreading `actions.search()` so lets use that instead to keep it DRY

src/core/actions.ts
```diff
@@ -45,6 +46,30 @@ namespace Actions {
...
+
+  // batch actions
+  // tslint:disable-next-line max-line-length
+  export type SwitchRefinement = [Actions.ResetPage, Actions.ResetPage, Actions.ResetRefinements, Actions.ResetPage, Actions.AddRefinement];
+  export type Search = [Actions.ResetPage, Actions.ResetPage, Actions.ResetRefinements, Actions.ResetPage, Actions.AddRefinement];
+  // tslint:disable-next-line max-line-length
+  export type ResetRecall = [Actions.ResetPage, Actions.ResetPage, Actions.ResetRefinements, Actions.ResetPage, Actions.UpdateQuery] |
+    [Actions.ResetPage, Actions.ResetPage, Actions.ResetRefinements, Actions.ResetPage, Actions.UpdateQuery, Actions.ResetPage, Actions.SelectRefinement];
+    // tslint:disable-next-line max-line-length
+  // export type UpdateSearch = Array<Actions.ResetPage | Actions.UpdateQuery | Actions.ResetRefinements | Actions.SelectRefinement | Actions.AddRefinement>;
```
effervescentia: we should uncomment this type and remove the other `type UpdateSearch` defined in this file

src/core/reducers/data/navigations.ts
```diff
@@ -105,20 +95,23 @@ export const deselectRefinement = (state: State, { navigationId, index: refineme
...
-    byId: {
-      ...state.byId,
-      [navigationId]: {
-        ...state.byId[navigationId],
-        selected: [refinementIndex],
-      },
-    },
-  };
-};
+const generateNavigationId = (state: State, navigationId: string, refinement: any, index: number) => ({
```
effervescentia: I think this would be better named `generateNavigation()` as that is what it does, rather than generate a `navigationId`

test/unit/core/reducers/data/navigations.ts
```diff
@@ -336,16 +287,16 @@ suite('navigations', ({ expect }) => {
       expect(reducer).to.eql(newState);
     });
 
-    it('should return state on UPDATE_SEARCH if no navigationId and payload clear is falsy', () => {
-      const reducer = navigations(state, {
-        type: Actions.UPDATE_SEARCH,
-        payload: {}
-      });
+    // it('should return state on UPDATE_SEARCH if no navigationId and payload clear is falsy', () => {
```
effervescentia: should this be commented out? if so, remove it

PR #70, opened 2017-09-21
fix(BBD-723): Fix indicies not being checked in shouldResetRefinements

Filip-F:
In the previous PR we had forgotten to cover the case where an index was passed in instead of a refinement, but there were no tests covering this case, so I added i and the corresponding tests.

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -48,14 +48,8 @@ export const handleError = (errorAction: Actions.Action<any>,  createAction: () 
 export const shouldResetRefinements =  ({ low, high, value, navigationId, range, index }: Actions.Payload.Search,
                                         state: any): boolean => {
   const currentRefinements = Selectors.selectedRefinements(state);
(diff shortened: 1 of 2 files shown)


src/core/utils.ts
```diff
@@ -45,11 +45,17 @@ export const refinementPayload = (field: string, valueOrLow: any, high: any = nu
...
+  let refinement = { low, high, value };
   // assumes only one refinement can be added at once
-  return !navigationId || currentRefinements.length !== 1 ||
-    !SearchAdapter.refinementsMatch(<any>refinement, currentRefinements[0], range ? 'Range' : 'Value');
+  if (!navigationId || currentRefinements.length !== 1) {
+    return true;
+  }
+
+  if (index) {
+    refinement = Selectors.refinementCrumb(state, navigationId, index);
```
effervescentia: if you have an `index` you can just return `!Selectors.isRefinementSelected(state, navigationId, index)`

PR #85, opened 2017-10-10
feat(BBD 766): Add location to recommendation config

Filip-F:
There is a corresponding PR to core for these config changes. https://github.com/groupby/storefront-core/pull/106

### DIFF
--- a/src/core/adapters/recommendations.ts
+++ b/src/core/adapters/recommendations.ts
@@ -65,31 +65,37 @@ namespace Recommendations {
       transform({ results, navigations, config }), availableNavigations);
   };
 
-  export const addLocationMatchExact = (request, state, config: Configuration) => {
+  export const addLocationToRequest = (request, state, config: Configuration):
+  RecommendationsBody | RecommendationsRequest => {
     const locationConfig = ConfigurationAdapter.extractLocation(config);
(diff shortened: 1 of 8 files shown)


src/core/configuration.ts
```diff
@@ -217,12 +213,22 @@ namespace Configuration {
...
+     */
+    location: Configuration.Recommendations.Location;
     productSuggestions: Configuration.Recommendations.ProductSuggestions;
     iNav: Configuration.Recommendations.INav;
     pastPurchases: Configuration.Recommendations.PastPurchases;
   }
 
   export namespace Recommendations {
+    export interface Location {
+      minSize: number;
```
effervescentia: I believe @Vic-Dev had a comment about `minSize` in the documentation pull request

src/core/adapters/recommendations.ts
```diff
@@ -64,6 +65,42 @@ namespace Recommendations {
       transform({ results, navigations, config }), availableNavigations);
   };
 
+  export const addLocationToRequest = (request, state, config: Configuration):
+  RecommendationsBody | RecommendationsRequest => {
```
effervescentia: even if it causes the line to go long, I prefer to keep types always to the right of the associated token(s) for clarity

PR #94, opened 2017-10-12
feat(BBD-782): Simplify action creators to improve testing and reuse

Vic-Dev:


src/flux-capacitor.ts
```diff
@@ -2,8 +2,9 @@ import { EventEmitter } from 'eventemitter3';
 import { BrowserBridge, Results } from 'groupby-api';
 import { Action as ReduxAction, Store as ReduxStore } from 'redux';
 import { Sayt } from 'sayt';
-import createActions from './core/action-creator';
+import wrapActionCreators from './core/actions';
```
Vic-Dev: imported but not used?

tslint-hound.json
```diff
@@ -0,0 +1,6 @@
+{
```
Vic-Dev: Should this be here?

PR #111, opened 2017-10-17
feat(BBD-789): UpdateCurrentPage action doesn't do any validation to make sure page exists

zfcb:
I moved the type checking to isValidPage instead of isDifferentPage

### DIFF
--- a/src/core/actions/validators.ts
+++ b/src/core/actions/validators.ts
@@ -102,12 +102,12 @@ export const isDifferentPageSize: Validator<number> = {
 
 export const isOnDifferentPage: Validator<number> = {
   func: (page, state) => Selectors.page(state) !== page,
-  msg: 'page size is already selected'
+  msg: 'page is already selected'
 };
 
 export const isValidPage: Validator<number> = {
-  func: (page, state) => typeof page === 'number' && Selectors.totalPages(state) >= page && page >= 1,
-  msg: 'page size is invalid'
+  func: (page, state) => typeof page === 'number' && page <= Selectors.totalPages(state) && page >= 1,
+  msg: 'page is invalid'
 };
 
 export const isDifferentAutocompleteQuery: Validator<string> = {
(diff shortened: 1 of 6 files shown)


src/core/selectors.ts
```diff
@@ -79,6 +79,11 @@ namespace Selectors {
     state.data.present.page.current;
 
   /**
+   * Returns the last page index, ie number of pages
+   */
+  export const pageAmount = (state: Store.State) =>
```
effervescentia: I think this would be better named `totalPages`

src/core/actions/validators.ts
```diff
@@ -101,10 +101,15 @@ export const isDifferentPageSize: Validator<number> = {
...
 
 export const isOnDifferentPage: Validator<number> = {
-  func: (page, state) => page !== null && Selectors.page(state) !== page,
+  func: (page, state) => Selectors.page(state) !== page,
   msg: 'page size is already selected'
 };
 
+export const isValidPage: Validator<number> = {
+  func: (page, state) => typeof page === 'number' && Selectors.totalPages(state) >= page && page >= 1,
+  msg: 'page size is invalid'
```
effervescentia: this message should be `page is invalid`

src/core/actions/validators.ts
```diff
@@ -101,10 +101,15 @@ export const isDifferentPageSize: Validator<number> = {
 };
 
 export const isOnDifferentPage: Validator<number> = {
-  func: (page, state) => page !== null && Selectors.page(state) !== page,
+  func: (page, state) => Selectors.page(state) !== page,
   msg: 'page size is already selected'
 };
 
+export const isValidPage: Validator<number> = {
+  func: (page, state) => typeof page === 'number' && Selectors.totalPages(state) >= page && page >= 1,
```
effervescentia: I would prefer `page <= Selectors.totalPages(state)` as `page` is the subject of the validation

PR #112, opened 2017-10-17
fix(BBD-782): middleware to handle thunks from within sagas

effervescentia:


src/core/store/middleware.ts
```diff
@@ -96,11 +97,12 @@ export namespace Middleware {
...
     return compose(
-      thunkEvaluator,
-      saveStateAnalyzer,
+      applyMiddleware(thunkEvaluator, saveStateAnalyzer),
       reduxBatch,
       applyMiddleware(...middleware),
-      reduxBatch
+      reduxBatch,
+      applyMiddleware(thunkEvaluator),
+      reduxBatch,
```
Filip-F: Why do we need to call reduxBatch, then thunk, then reduxBatch? If the action is a batchAction, then it'll go through the whole middleware stack then, wouldn't it? Would it not suffice to call thunkEvaluator then reduxBatch?

PR #114, opened 2017-10-18
docs(BBD-780): Action creators documentation

Vic-Dev:


effervescentia:
make sure that when you merge it you use the `docs()` commit type not `feat()`

PR #119, opened 2017-10-31
feat(BBD-822): Store the record index for each product

Vic-Dev:


src/core/adapters/search.ts
```diff
@@ -144,10 +144,26 @@ namespace Adapter {
...
+    return results.records.map(({ allMeta }) => {
+      const meta: any = {};
+
+      if (allMeta[state.session.config.recommendations.idField] in pastPurchases) {
+        meta.pastPurchase = true;
+      }
+
+      return {
+        meta,
+        index: startIndex++,
```
effervescentia: should be able to use the index from `map()` and add it to `startIndex`
make `startIndex` a `const`

src/core/selectors.ts
```diff
@@ -113,23 +113,13 @@ namespace Selectors {
    * Returns the current products
    */
   export const products = (state: Store.State) =>
-    state.data.present.products;
+    Search.extractData(state.data.present.products);
 
   /**
    * Returns the current products extended with metadata
    */
-  export const productsWithMetadata = (state: Store.State, idField: string) => {
```
effervescentia: resurrect this selector as `productsWithPastPurchase`

src/core/adapters/search.ts
```diff
@@ -144,10 +144,26 @@ namespace Adapter {
...
+    products.map(({ data }) => data);
 
-  export const extractProduct = ({ allMeta }) => allMeta;
+  export const extractProducts = (state: Store.State, results: Results) => {
+    const pastPurchases = Selectors.pastPurchaseProductsBySku(state);
+    let startIndex = results.pageInfo.recordStart;
+    return results.records.map(({ allMeta }) => {
+      const meta: any = {};
+
+      if (allMeta[state.session.config.recommendations.idField] in pastPurchases) {
```
effervescentia: we should only add this when selecting the data back out to ensure we never get out of sync

test/unit/core/adapters/search.ts
```diff
@@ -335,28 +335,48 @@ suite('Search Adapter', ({ expect, stub }) => {
     });
   });
 
+  describe('extractdata()', () => {
```
effervescentia: `extractData()`

src/core/adapters/search.ts
Vic-Dev: what?

src/core/adapters/search.ts
```diff
@@ -144,10 +144,21 @@ namespace Adapter {
...
-    results.records.map(Adapter.extractProduct);
-
-  export const extractProduct = ({ allMeta }) => allMeta;
+  export const extractData = (products: Store.ProductWithMetadata[]) =>
+    products.map(({ data }) => data);
+
+  export const augmentProducts = (results: Results) => {
+    const startIndex = results.pageInfo.recordStart;
+    return results.records.map(({ allMeta }, index) => {
+      const meta: any = {};
```
effervescentia: could probably just inline this if we don't plan on mutating it

effervescentia: PS can you extract `collection` from the record and put that in the `meta`?

src/core/observer.ts
```diff
@@ -137,7 +139,10 @@ namespace Observer {
           },
 
           recommendations: {
-            products: emit(Events.RECOMMENDATIONS_PRODUCTS_UPDATED)
+            // products: emit(Events.RECOMMENDATIONS_PRODUCTS_UPDATED)
```
effervescentia: can we remove this comment?

PR #123, opened 2017-11-02
feat(BBD-808): RTB - Bias results in real time

zfcb:


src/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,74 @@
...
+    allIds = allIds.slice(0, config.globalMaxBiases);
+  }
+
+  return { ...state, biasing: { ...state.biasing, byId, allIds }};
+};
+
+// tslint:disable-next-line max-line-length
+export const insertSorted = (allIds: Store.Personalization.BiasKey[], { variant, key }: Store.Personalization.BiasKey) => {
+  const noDuplicate = allIds.filter((id) => !(id.variant === variant && id.key === key));
+  return [{ variant, key }, ...noDuplicate]; //TODO: insert sorted
```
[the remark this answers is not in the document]
Filip-F: Not going to fix this one, for the mvp we're just inserting it at the start of the array, but in the future we're going to insert it in its proper sorted location in the array (which may not necessarily be at the front if we're using frequency).

test/unit/core/store/index.ts
```diff
@@ -24,6 +25,9 @@ suite('Store', ({ expect, spy, stub }) => {
       const createSagaMiddleware = stub(reduxSaga, 'default').returns(sagaMiddleware);
       const createMiddleware = stub(Middleware, 'create').returns(middleware);
       const createSagas = stub(sagas, 'default').returns(['m', 'n', 'o']);
+      const isRealTimeBiasEnabled = stub(Adapter, 'isRealTimeBiasEnabled').returns(false);
+      // const persistStore = stub(persist, 'persistStore');
+      // TODO: make this work
```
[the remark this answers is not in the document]
Filip-F: Leaving this in since this is something we should actually test.

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+    return {
+      expiry: state.expiry,
+      allIds,
+      byId
+    };
+  };
+
+  export const convertBiasToSearch = (state: Store.State) => {
+    // const byId = Selectors.realTimeBiasesById(state);
```
effervescentia: remove comment

src/core/sagas/product-details.ts
```diff
@@ -29,6 +29,7 @@ export namespace Tasks {
         throw new Error(`no records found matching id: ${id}`);
       }
     } catch (e) {
+      console.error(e, 'rr');
```
effervescentia: remove logging

src/core/store/middleware.ts
```diff
@@ -80,6 +88,19 @@ export namespace Middleware {
     };
   }
 
+  export function personalizationAnalyzer(store: Store<any>) {
+    return (next) => (action) => {
+      if (ConfigurationAdapter.isRealTimeBiasEnabled(Selectors.config(store.getState())) &&
```
effervescentia: should just call `store.getState()` once and save the result for reuse

src/core/store/middleware.ts
```diff
@@ -5,7 +5,11 @@ import { ActionCreators } from 'redux-undo';
 import * as validatorMiddleware from 'redux-validator';
 import FluxCapacitor from '../../flux-capacitor';
 import Actions from '../actions';
+import Creators from '../actions/creators';
```
effervescentia: can we import this as `ActionCreators` just for clarity

src/core/store/middleware.ts
```diff
@@ -102,7 +124,7 @@ export namespace Middleware {
       reduxBatch,
       applyMiddleware(...middleware),
       reduxBatch,
-      applyMiddleware(thunkEvaluator),
+      applyMiddleware(thunkEvaluator, Middleware.validator()),
```
effervescentia: if we have two `Middleware.validator()`s we can just call it once and save the result for reuse

src/core/store/middleware.ts
zfcb: Conflicts with ActionCreators from redux-undo

src/core/adapters/personalization.ts
Filip-F: That was left in since we might use the actual data on the biases at some point (if we ever decide to do decay for example), so I figured I'd leave that in but commented out. Right now all that we care is that the bias exists, so no need to have it for now.

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+      stub(Selectors, 'config').returns(config);
+    });
+
+    it('should generate new timeStamp if variant and key do not exist', () => {
+      const byId = {};
+      stub(Selectors, 'realTimeBiasesById').returns(byId);
+      stub(Selectors, 'refinementCrumb').returns(refinement);
+      stub(Adapter, 'generateNewBias').returns({ lastUsed: 11111 });
+
+      const bias = {
```
effervescentia: `const` above `stub()`s

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
+import Adapter from '../../../../src/core/adapters/personalization';
+import Selectors from '../../../../src/core/selectors';
+import suite from '../../_suite';
+
+suite('Personalization Adapter', ({ expect, stub }) => {
+
+  describe('extractBias()', () => {
+    let config = {
```
effervescentia: do these need to be `let`s?

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+      );
+      expect(result).to.deep.equal(bias);
+      expect(Adapter.generateNewBias).to.have.been.called;
+
+    });
+
+    it('should overwrite the timestamp if variant and key already exist', () => {
+      const byId = { color: { blue: { lastUsed: 11111 }}};
+      stub(Selectors, 'realTimeBiasesById').returns(byId);
+      stub(Selectors, 'refinementCrumb').returns(refinement);
```
effervescentia: `const`s above `stub()`s

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+      const s = stub(Date, 'now').returns(20000000);
+      const bias = {
+        variant: refinement.field,
+        key: refinement.value,
+        bias: {
+          lastUsed: 20000
+        }
+      };
+
+      const result = Adapter.extractBias(action, store);
```
effervescentia: newline between action and expectations

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+      expect(result).to.deep.equal(bias);
+      expect(Adapter.generateNewBias).to.not.have.been.called;
+    });
+
+    it('should generate new timestamp if variant exist and key does not exist', () => {
+      const byId = { color: { red: { lastUsed: 11111 }}};
+      stub(Selectors, 'realTimeBiasesById').returns(byId);
+      stub(Selectors, 'refinementCrumb').returns(refinement);
```
effervescentia: `const`s above `stub()`s

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+      stub(Adapter, 'generateNewBias').returns({ lastUsed: 22222 });
+      const bias = {
+        variant: refinement.field,
+        key: refinement.value,
+        bias: {
+          lastUsed: 22222
+        }
+      };
+
+      const result = Adapter.extractBias(action, store);
```
effervescentia: newline between action and expectation

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+    it('should return null if the attribute is not in the config', () => {
+      const newRefinement = {
+        field: 'brand',
+        value: 'Nike'
+      };
+      const byId = {};
+      stub(Selectors, 'realTimeBiasesById').returns(byId);
+      stub(Selectors, 'refinementCrumb').returns(newRefinement);
+
+      const result = Adapter.extractBias(action, store);
```
effervescentia: newline between action and expectation

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+      expect(result).to.deep.equal(null);
+    });
+  });
+
+  describe('generateNewBias()', () => {
+
```
effervescentia: no newline here

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+    it('should generate new bias', () => {
+      stub(Date, 'now').returns(1000000);
+
+      const result = Adapter.generateNewBias();
```
effervescentia: newline between action and expectation

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+      expect(result).to.deep.equal({ lastUsed: 1000000 / 1000 });
+    });
+  });
+
+  describe('transformToBrowser()', () => {
+
+    it('should return empty array if the state does not have allIds', () => {
+      const state: any = {
```
effervescentia: can be one line

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+        expiry: 2000
+      };
+      const key = 'test';
+      const result = Adapter.transformToBrowser(state, key);
```
effervescentia: newline before and after

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+        byId: {
+          color: {
+            blue: {
+              lastUsed: 11111
+            }
+          }
+        }
+      };
+      const key = 'test';
+      const result = Adapter.transformToBrowser(state, key);
```
effervescentia: newline before and after

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+      const now = Date.now() / 1000;
+      const browserStorage = {
+        expiry: oneDayInSec * 30,
+        allIds: [{ variant: 'color', key: 'blue', lastUsed: now - oneDayInSec * 5 },
+        { variant: 'color', key: 'red', lastUsed: now - oneDayInSec * 21 },
+        { variant: 'brand', key: 'Nike', lastUsed: now - oneDayInSec * 33 }]
+      };
+
+      const result = Adapter.transformFromBrowser(browserStorage, key);
+      const biasFromBrowser = {
```
effervescentia: `const` should be declared at top of `it()`

src/core/store/middleware.ts
effervescentia: then we can rename the `redux-undo` one

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,227 @@
...
+                strength: 'Absolute_Increase'
+              }
+            }
+          }
+        }
+      };
+      stub(Selectors, 'realTimeBiasesAllIds').returns(allIds);
+      stub(Selectors, 'config').returns(config);
+
+      const result = Adapter.convertBiasToSearch(state);
```
effervescentia: newline after action

test/unit/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,190 @@
+import { Actions, Store } from '../../../../../src/core';
+import * as personalization from '../../../../../src/core/reducers/data/personalization';
+import suite from '../../../_suite';
+
+suite('personalization', ({ expect, spy, stub }) => {
+  describe('updatePersonalization()', () => {
+    const state: Store.Personalization = {
+      biasing: <any>{ a: 1 }
+    };
+    const updatePersonalization = personalization.default;
```
effervescentia: take a look at the other reducer test for the pattern of how to import and name this, 
should use the `import` syntax to get both:
```
import reducer, * as personalization from '../wherever';
```

test/unit/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,190 @@
...
+          },
+        },
+        allIds: [
+          { variant: 'a', key: 'b' },
+          { variant: 'c', key: 'd' },
+        ],
+        expiry: 2
+      }
+    };
+
```
effervescentia: no newline here

test/unit/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,190 @@
...
+        bias: 3,
+        config: {
+          attributes: {
+            a: { maxBiases: 0 }
+          }
+        }
+      };
+      const allIds = [0, 1, 2, 3, { variant: 'a' }, 4];
+      stub(personalization, 'insertSorted').returns(allIds);
+      const removeLast = stub(personalization, 'removeLast').returns(allIds);
```
effervescentia: `const`s above `stub()`s

test/unit/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,190 @@
...
+        key: 'z',
+        bias: 3,
+        config
+      };
+      const slice = spy();
+      const allIds = {
+        length: 10020000,
+        slice
+      };
+      stub(personalization, 'insertSorted').returns(allIds);
```
effervescentia: `const`s above `stub()`s

test/unit/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,190 @@
...
+      expect(ret).to.eql(newArr);
+    });
+
+    it('shoud not modify input array if variant does not exists in array', () => {
+      const ret = personalization.removeLast(arr, 'd');
+
+      expect(ret).to.eql(arr);
+    });
+  });
+  describe('insertSorted', () => {
```
effervescentia: newline between `describe()` blocks

test/unit/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,190 @@
...
+    const arr = [
+      { variant: 'a', key: 1},
+      { variant: 'b', key: 2},
+      { variant: 'c', key: 3},
+    ];
+
+    it('shoud insert element to front of array', () => {
```
effervescentia: `'should'`

test/unit/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,190 @@
...
+      const obj = { variant: 'd', key: 4};
+      const newArr = [
```
effervescentia: can be one line

test/unit/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    it('shoud insert element to front of array and remove duplicates', () => {
+      const obj = { variant: 'b', key: 2};
+      const newArr = [
+        { variant: 'b', key: 2 },
+        { variant: 'a', key: 1 },
+        { variant: 'c', key: 3 },
+      ];
+
+      const ret = personalization.insertSorted(<any>arr, <any>obj);
```
effervescentia: rather than cast to `<any>` here, why not set the type of `arr` to be `any[]` and `obj` to be `any` when initialized

test/unit/core/sagas/products.ts
```diff
@@ -421,14 +422,19 @@ suite('products saga', ({ expect, spy, stub }) => {
         const payload = { a: 'b' };
         const action: any = { payload };
         const receiveProductsAction: any = { c: 'd' };
-        const request = { e: 'f' };
+        const biases = [];
+        const request = { e: 'f',
```
effervescentia: make one line

test/unit/core/selectors.ts
```diff
@@ -524,6 +525,130 @@ suite('selectors', ({ expect, stub }) => {
     });
   });
 
+  describe('realTimeBiasesById()', () => {
+    it('should return byId', () => {
+      const allIds = [1,2,3,4,5];
```
effervescentia: space between array items

test/unit/core/selectors.ts
```diff
@@ -524,6 +525,130 @@ suite('selectors', ({ expect, stub }) => {
...
+      const byId = {
```
effervescentia: can be one line

test/unit/core/selectors.ts
```diff
@@ -524,6 +525,130 @@ suite('selectors', ({ expect, stub }) => {
...
+        }
+      };
+
+      expect(Selectors.realTimeBiasesById(<any>state)).to.eq(byId);
+    });
+  });
+
+  describe('realTimeBiasesAllIds()', () => {
+    it('should return allIds', () => {
+      const allIds = [1,2,3,4,5];
```
effervescentia: space between array items

test/unit/core/selectors.ts
effervescentia: can be reused across `describe()`s

test/unit/core/selectors.ts
effervescentia: can be reused across `describe()`s

test/unit/core/selectors.ts
```diff
@@ -524,6 +525,130 @@ suite('selectors', ({ expect, stub }) => {
...
+      const byId = {
```
effervescentia: one line

test/unit/core/selectors.ts
```diff
@@ -524,6 +525,130 @@ suite('selectors', ({ expect, stub }) => {
...
+                rehydrated: true
+              }
+            }
+          }
+        }
+      };
+      const isRealTimeBiasEnabled = stub(ConfigurationAdapter, 'isRealTimeBiasEnabled').returns(false);
+      const config = stub(Selectors, 'config');
+
+      expect(Selectors.realTimeBiasesHydrated(<any>state)).to.be.true;
```
effervescentia: can set type of `state` to `any` when initialized

test/unit/core/store/index.ts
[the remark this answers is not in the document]
effervescentia: still need to "make this work"?
oh this is for the testing
in that case just remove it or wrap the call to `persistStore` in a function so it can be evaluated after module load

test/unit/core/store/index.ts
```diff
@@ -36,6 +40,29 @@ suite('Store', ({ expect, spy, stub }) => {
         .and.calledWith('m')
         .and.calledWith('n')
         .and.calledWith('o');
+      // expect(persistStore).to.be.called;
+    });
+
+    it('should not persist store if real time bias disabled', () => {
+      // tslint:disable-next-line variable-name
+      const __config = { a: 'b' };
```
effervescentia: seems like a lot of these `const`s are not being used in this `it()`

test/unit/core/store/middleware.ts
```diff
@@ -4,10 +4,15 @@ import reduxLogger from 'redux-logger';
 import { ActionCreators } from 'redux-undo';
 import * as sinon from 'sinon';
 import Actions from '../../../../src/core/actions';
+import Creators from '../../../../src/core/actions/creators';
```
effervescentia: also call `ActionCreators` and rename the type from `redux-undo`

test/unit/core/store/middleware.ts
```diff
@@ -218,4 +224,61 @@ suite('Middleware', ({ expect, spy, stub }) => {
       expect(thunk).to.be.calledWithExactly(state);
     });
   });
+
+  describe('personalizationAnalyzer()', () => {
```
effervescentia: try to match up the object types here `conf` should be an `object`, `state` should be an `object`

effervescentia: can also probably be reused across multiple `it()`s

test/unit/flux-capacitor.ts
```diff
@@ -279,6 +280,34 @@ suite('FluxCapacitor', ({ expect, spy, stub }) => {
...
     });
+
+    describe('awaitBiasingRehydration()', () => {
+      it('should dispatch given action if data loaded from browser', () => {
+        const action = 'action';
+        const getState = store.getState = spy();
+        const dispatch = store.dispatch = spy();
+        stub(Selectors, 'realTimeBiasesHydrated').returns(true);
+
+        flux.awaitBiasingRehydration(<any>action);
```
effervescentia: newline between actions and expectations

test/unit/flux-capacitor.ts
```diff
@@ -279,6 +280,34 @@ suite('FluxCapacitor', ({ expect, spy, stub }) => {
...
+      });
+
+      it('should wait on personalization biasing rehydrated if data not loaded from browser', () => {
+        const action = 'action';
+        const realtimeBiasesHydrated = stub(Selectors, 'realTimeBiasesHydrated').returns(false);
+        const once = flux.once = spy();
+        const dispatch = store.dispatch = spy();
+        store.getState = spy();
+
+        flux.awaitBiasingRehydration(<any>action);
```
effervescentia: newline between actions and expectations

effervescentia: can set `any` type on `action` at initialization

test/unit/flux-capacitor.ts
```diff
@@ -279,6 +280,34 @@ suite('FluxCapacitor', ({ expect, spy, stub }) => {
...
+        expect(dispatch).to.not.be.called;
+        expect(once).to.be.calledWith(Events.PERSONALIZATION_BIASING_REHYDRATED);
+
```
effervescentia: no newline required here

test/unit/flux-capacitor.ts
```diff
@@ -279,6 +280,34 @@ suite('FluxCapacitor', ({ expect, spy, stub }) => {
...
+        once.getCall(0).args[1]();
```
effervescentia: rather than doing this, check the documentation for `sinon.match()` and look at how we use it elsewhere in the tests

src/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,73 @@
...
+};
+
+export const removeLast = (allIds, variant) => {
```
Vic-Dev: More appropriate name ```removeVariant```

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,79 @@
...
+    lastUsed: Math.floor(Date.now() / 1000)
+  });
+
+  // tslint:disable-next-line max-line-length
+  export const transformToBrowser = (state: Store.Personalization.Biasing, reducerKey: string): BrowserStorageState => ({
+    expiry: state.expiry,
+    allIds: state.allIds ? state.allIds.map(({ variant, key }) => ({
+      variant,
+      key,
+      ...state.byId[variant][key]
```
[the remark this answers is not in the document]
zfcb: It shouldn't, since byId[variant][key] does not store variant and key

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,79 @@
...
+  };
+
+  export const convertBiasToSearch = (state: Store.State) => {
+    const allIds = Selectors.realTimeBiasesAllIds(state);
+    const config = Selectors.config(state).personalization.realTimeBiasing;
+
+    return allIds.map(({ variant, key }) => ({
+      name: variant,
+      content: key,
+      strength: (config.attributes[variant] && config.attributes[variant].strength) || config.strength
```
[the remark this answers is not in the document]
zfcb: In the case that strength is not defined for a variant, we need to fall back to config.strength. I don't think a ternary, i.e

```js
config.attributes[variant] ? config.attributes[variant].strength : config.strength
```

would work here. modifications to make it work would end up making it messier than the code without ternary above.

src/core/observer.ts
```diff
@@ -155,6 +155,16 @@ namespace Observer {
           sorts: emit(Events.SORTS_UPDATED),
 
           template: emit(Events.TEMPLATE_UPDATED),
+
+          personalization: {
+            _persist: {
+              rehydrated: (oldState, newState, path) => {
+                if (newState && newState !== oldState) {
```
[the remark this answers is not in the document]
zfcb: Interesting, thanks.

test/unit/core/actions/validators.ts
```diff
@@ -344,6 +344,47 @@ suite('validators', ({ expect, spy, stub }) => {
     });
   });
 
+  describe('isValidBias', () => {
+    it('should be invalid valid if payload is falsy', () => {
```
Vic-Dev: ```invalid valid``` what?

test/unit/core/actions/validators.ts
```diff
@@ -344,6 +344,47 @@ suite('validators', ({ expect, spy, stub }) => {
...
+      const payload = false;
+
+      expect(validators.isValidBias.func(<any>payload)).to.be.false;
+    });
+
+    it('should be invalid valid if payload.key is falsy and payload.variant is falsy', () => {
```
Vic-Dev: again

test/unit/core/actions/validators.ts
```diff
@@ -344,6 +344,47 @@ suite('validators', ({ expect, spy, stub }) => {
...
+      const payload = {};
+
+      expect(validators.isValidBias.func(<any>payload)).to.be.false;
+    });
+
+    it('should be invalid valid if payload.key is falsy', () => {
```
Vic-Dev: again

test/unit/core/actions/validators.ts
```diff
@@ -344,6 +344,47 @@ suite('validators', ({ expect, spy, stub }) => {
...
+      const payload = {
+        variant: 'asdf',
+        key: false
+      };
+
+      expect(validators.isValidBias.func(<any>payload)).to.be.false;
+    });
+
+    it('should be invalid valid if payload.variant is falsy', () => {
```
Vic-Dev: again

src/core/store/middleware.ts
```diff
@@ -34,8 +38,12 @@ export const SEARCH_CHANGE_ACTIONS = [
   Actions.UPDATE_CURRENT_PAGE,
 ];
 
+export const PERSONALIZATION_CHANGE_ACTIONS = [
+  Actions.SELECT_REFINEMENT,
```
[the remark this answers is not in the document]
zfcb: TODO, may need to also change the middleware

src/flux-capacitor.ts
```diff
@@ -130,6 +130,14 @@ class FluxCapacitor extends EventEmitter {
     this.store.dispatch(this.actions.receiveAutocompleteProductRecords(
       this.selectors.queryPastPurchases(this.store.getState())));
   }
+  
+  awaitBiasingRehydration(action: Actions.Action) {
```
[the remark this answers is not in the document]
Filip-F: It's used in core, where would you prefer this defined then?

src/core/reducers/data/personalization.ts
Filip-F: How about something like `removeLastForVariant`. I agree that `removeLast` is not a good name, but what the function does is it finds the item for that variant furthest back in the array and removes it. It doesn't remove all items of that variant.

Vic-Dev: Oh I didn't even notice the for loop was going backwards lol.

Vic-Dev: Maybe removeOldest?

test/unit/flux-capacitor.ts
zfcb: Are you referring to the calledWith? If so then wouldn't an exact argument rather than sinon.match be better?

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,79 @@
...
+    const olderThanTime = Math.floor(Date.now() / 1000) - state.expiry;
+    const filteredState = state.allIds.filter((element) => element.lastUsed >= olderThanTime);
+    let allIds = [];
+    let byId = {};
+    filteredState.forEach(({ variant, key, lastUsed }) => {
+      allIds.push({ variant, key });
+      if (!byId[variant]) {
+        byId[variant] = {};
+      }
+      byId[variant][key] = { lastUsed };
```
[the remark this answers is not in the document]
zfcb: Good catch thanks

src/core/store/middleware.ts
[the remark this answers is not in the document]
zfcb: Addressed

src/core/adapters/personalization.ts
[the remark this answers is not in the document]
Vic-Dev: I think Ben meant 
```js
config.attributes[variant] && config.attributes[variant].strength ? config.attributes[variant].strength : config.strength
```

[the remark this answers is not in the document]
Filip-F: @effervescentia confirm? Personally I find the one already there clearer, but if you'd prefer we can use this.

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  // tslint:disable-next-line max-line-length
+  export const extractRefinement = ({ type, payload }: ExtractableAction, state: Store.State): { field: string, value: string } => {
+    // tslint:disable-next-line one-variable-per-declaration
+    let field = undefined, value = undefined;
+
+    switch (type) {
+      case Actions.ADD_REFINEMENT:
+        if (payload.range) {
```
effervescentia: rather than `break` early, why not just wrap the logic in the negation?
```
  case Actions.ADD_REFINEMENT:
    if (!payload.range) {
      field = payload.navigationId;
      value = payload.value;
    }
    break;
```

This reads more cleanly as we only have one `break`

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+      key: value,
+      bias: (byId[field] && byId[field][value]) ? {
+        ...byId[field][value],
+        lastUsed: Math.floor(Date.now() / 1000)
+      } : generateNewBias()
+    };
+  };
+
+  // tslint:disable-next-line max-line-length
+  export const extractRefinement = ({ type, payload }: ExtractableAction, state: Store.State): { field: string, value: string } => {
```
effervescentia: rather than have this function set `field` and `value` then return them at the end, why not just return them from within your switch statement?

```js
switch (type) {
  case Actions.ADD_REFINEMENT:
    if (payload.range) {
      return { field: payload.navigationId, value: payload.value };
    }
  case Actions.SELECT_REFINEMENT:
    return Selectors.refinementCrumb(state, payload.navigationId, payload.index);
}

return {};
```

effervescentia: also does the `type` param get used?

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+        field = refinement.field;
+        value = refinement.value;
+        break;
+      default:
+    }
+
+    return { field, value };
+  };
+
+  export const generateNewBias = () => ({
```
effervescentia: this can be a one-line function

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+    lastUsed: Math.floor(Date.now() / 1000)
+  });
+
+  // tslint:disable-next-line max-line-length
+  export const transformToBrowser = (state: Store.Personalization.Biasing, reducerKey: string): BrowserStorageState => ({
+    allIds: state.allIds ? state.allIds.map(({ variant, key }) => ({
```
effervescentia: will `allIds` ever be falsy?

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+      variant,
+      key,
+      ...state.byId[variant][key]
```
effervescentia: do we care that the values of `variant` or `key` might be overridden by this spread?

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+    })) : []
+  });
+
+  // tslint:disable-next-line max-line-length
+  export const transformFromBrowser = (incomingState: BrowserStorageState, state: Store.State): Store.Personalization.Biasing => {
+    const config = Selectors.config(state);
+    const olderThanTime = Math.floor(Date.now() / 1000) - DAYS_IN_SECONDS * ConfigAdapter.extractRealTimeBiasingExpiry(config);
```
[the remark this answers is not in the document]
effervescentia: ^^^ add `tstlint` disabling comment

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+    };
+  };
+
+  export const convertBiasToSearch = (state: Store.State) => {
+    const allIds = Selectors.realTimeBiasesAllIds(state);
+    const config = Selectors.config(state).personalization.realTimeBiasing;
+    const selectedRefinements = Selectors.selectedRefinements(state);
+
+    return allIds.filter(({ variant, key }) => {
+      return !selectedRefinements.find(({ navigationName, type, value }) => {
```
effervescentia: can use `some()` here instead of `find()`

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+      byId
+    };
+  };
+
+  export const convertBiasToSearch = (state: Store.State) => {
+    const allIds = Selectors.realTimeBiasesAllIds(state);
+    const config = Selectors.config(state).personalization.realTimeBiasing;
+    const selectedRefinements = Selectors.selectedRefinements(state);
+
+    return allIds.filter(({ variant, key }) => {
```
effervescentia: don't need a function body for the one line return

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+      return !selectedRefinements.find(({ navigationName, type, value }) => {
+        return type === 'Value' && navigationName === variant && key && value === key;
+      });
+    }).map(({ variant, key }) => ({
+      name: variant,
+      content: key,
+      strength: (config.attributes[variant] && config.attributes[variant].strength) || config.strength
```
effervescentia: this would be more clear as a ternary statement

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+    }));
+  };
+
+  // tslint:disable-next-line max-line-length
+  export const pruneBiases = (allIds: Store.Personalization.BiasKey[], variant: string, variantCount: number, config: Configuration.Personalization.RealTimeBiasing) => {
+    if (config.attributes[variant] && (variantCount >= (config.attributes[variant].maxBiases))) {
```
effervescentia: two sets of unnecessary parentheses

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+      for (let i = allIds.length - 1; i >= 0; i--) {
+        if (allIds[i].variant === variant) {
```
effervescentia: can we use `findIndex()` here instead?

effervescentia: If you need to find the first one from the end, you can use `reverse()`

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+          return [...allIds.slice(0, i), ...allIds.slice(i + 1)];
```
effervescentia: why not use `splice()` here?

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,126 @@
...
+        }
+      }
+    }
+    if (allIds.length > config.maxBiases) {
+      return allIds.slice(0, config.maxBiases);
```
effervescentia: is this in order to shallow clone the array? could be `[...config.maxBiases]` if you really need that

src/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,58 @@
...
+    ...state.biasing.byId,
+    [payload.variant]: {
+      ...state.biasing.byId[payload.variant],
+      [payload.key]: payload.bias
+    }
+  };
+
+  let allIds = insertSorted(state.biasing.allIds, { variant: payload.variant, key: payload.key });
+  const config = payload.config;
+  allIds = Adapter.pruneBiases(allIds, payload.variant, Object.keys(byId[payload.variant]).length, config);
```
effervescentia: I'd rather not re-assign this value right after initializing it, maybe return the value of `pruneBiases()` inline in the returned object literal

src/core/sagas/products.ts
```diff
@@ -102,11 +111,20 @@ export namespace Tasks {
       yield effects.put(<any>flux.actions.receiveMoreProducts(e));
     }
   }
+
+  export function* fetchProductsWhenHydrated(flux: FluxCapacitor, action: Actions.fetchProductsWhenHydrated) {
+    if (Selectors.realTimeBiasesHydrated(flux.store.getState())) {
+      flux.store.dispatch(action.payload);
+    } else {
+      flux.once(Events.PERSONALIZATION_BIASING_REHYDRATED, () => flux.store.dispatch(action.payload));
```
effervescentia: we don't need to access `flux` in order to do a dispatch here, you can use the redux effect `put()`

test/unit/core/adapters/configuration.ts
```diff
@@ -376,4 +376,24 @@ suite('Configuration Adapter', ({ expect, stub }) => {
...
+      // tslint:disable-next-line max-line-length
+      expect(Adapter.isRealTimeBiasEnabled(<any>{ personalization: { realTimeBiasing: { attributes: null } } })).to.be.false;
+    });
+  });
+
+  describe('extractRealTimeBiasingExpiry()', () => {
+    it('should return the expiry time for real time biasing', () => {
+      const expiry = 3;
+      // tslint:disable-next-line max-line-length
+      expect(Adapter.extractRealTimeBiasingExpiry(<any>{ personalization: { realTimeBiasing: { expiry } } })).to.eql(expiry);
```
effervescentia: should be able to do direct equality check `eq()`

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+      stub(Selectors, 'realTimeBiasesById').returns(byId);
+      stub(Adapter, 'extractRefinement').returns(refinement);
+      stub(Adapter, 'generateNewBias').returns({ lastUsed: 11111 });
+
+      const result = Adapter.extractBias(action, store);
+
+      expect(Selectors.config).to.be.calledWithExactly(store);
+      expect(Selectors.realTimeBiasesById).to.be.calledWithExactly(store);
+      expect(Adapter.extractRefinement).to.be.calledWithExactly(action, store);
+      expect(result).to.deep.equal(bias);
```
effervescentia: can just use `eql()`

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+      expect(Adapter.generateNewBias).to.have.been.called;
+    });
+
+    it('should overwrite the timestamp if variant and key already exist', () => {
+      const byId = { color: { blue: { lastUsed: 11111 } } };
+      const s = stub(Date, 'now').returns(20000000);
```
effervescentia: please choose a better name than `s`

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+          lastUsed: 20000
+        }
+      };
+      stub(Selectors, 'realTimeBiasesById').returns(byId);
+      stub(Adapter, 'extractRefinement').returns(refinement);
+      stub(Adapter, 'generateNewBias');
+
+      const result = Adapter.extractBias(action, store);
+
+      expect(result).to.deep.equal(bias);
```
effervescentia: can just use `eql()`

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+        }
+      };
+      stub(Selectors, 'realTimeBiasesById').returns(byId);
+      stub(Adapter, 'extractRefinement').returns(refinement);
+      stub(Adapter, 'generateNewBias').returns({ lastUsed: 22222 });
+
+      const result = Adapter.extractBias(action, store);
+
+      expect(Adapter.generateNewBias).to.have.been.called;
+      expect(result).to.deep.equal(bias);
```
effervescentia: can just use `eql()`

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+        field: 'brand',
+        value: 'Nike'
+      };
+      const byId = {};
+      stub(Selectors, 'realTimeBiasesById').returns(byId);
+      stub(Adapter, 'extractRefinement').returns(newRefinement);
+
+      const result = Adapter.extractBias(action, store);
+
+      expect(result).to.deep.equal(null);
```
effervescentia: can just expect `.to.be.null;`

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+      expect(Adapter.generateNewBias).to.have.been.called;
+      expect(result).to.deep.equal(bias);
+    });
+
+    it('should return null if the attribute is not in the config', () => {
+      const newRefinement = {
+        field: 'brand',
+        value: 'Nike'
+      };
+      const byId = {};
```
effervescentia: don't need this variable if it's only used in the `returns()`, just inline it

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+      stub(Adapter, 'generateNewBias').returns({ lastUsed: 22222 });
+
+      const result = Adapter.extractBias(action, store);
+
+      expect(Adapter.generateNewBias).to.have.been.called;
+      expect(result).to.deep.equal(bias);
+    });
+
+    it('should return null if the attribute is not in the config', () => {
+      const newRefinement = {
```
effervescentia: same as below

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+      const result = Adapter.generateNewBias();
+
+      expect(result).to.deep.equal({ lastUsed: 1000000 / 1000 });
+    });
+  });
+
+  describe('transformToBrowser()', () => {
+    it('should return empty array if the state does not have allIds', () => {
+      const state: any = { expiry: 2000 };
+      const key = 'test';
```
effervescentia: doesn't need to be a variable as it's only used once

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+        }],
+        byId: {
+          color: {
+            blue: {
+              lastUsed: 11111
+            }
+          }
+        }
+      };
+      const key = 'test';
```
effervescentia: doesn't need to be a variable

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+            },
+            red: {
+              lastUsed: now - oneDayInSec * 21
+            }
+          }
+        }
+      };
+      const realTimeBiasing = 'real';
+      const conf = { personalization: { realTimeBiasing } };
+      const state: any = 's';
```
effervescentia: if we're going to make a mock state, please at least make it the correct type: `{}`

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+      const newArr = [
+        { variant: 'a', key: 1 },
+        { variant: 'b', key: 2 },
+        { variant: 'c', key: 3 },
+      ];
+
+      const ret = Adapter.pruneBiases(<any>arr, 'b', 2, <any>{ attributes: {
+        b: {
+          maxBiases: 1
+        }}});
```
effervescentia: weird bracket spacing

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+    ];
+
+    it('shoud remove last element of a specific variant if it exists ', () => {
+      const newArr = [
+        { variant: 'a', key: 1 },
+        { variant: 'b', key: 2 },
+        { variant: 'c', key: 3 },
+      ];
+
+      const ret = Adapter.pruneBiases(<any>arr, 'b', 2, <any>{ attributes: {
```
effervescentia: rather than cast to `<any>` here, just set the type of `arr` to `any[]` at declaration

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+          }
+        }
+      };
+      stub(Selectors, 'realTimeBiasesAllIds').returns(allIds);
+      stub(Selectors, 'selectedRefinements').returns([]);
+      stub(Selectors, 'config').returns(config);
+
+      const result = Adapter.convertBiasToSearch(state);
+
+      expect(result).to.deep.equal([{
```
effervescentia: can just use `eql()`

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,380 @@
...
+      stub(Selectors, 'realTimeBiasesAllIds').returns(allIds);
+      stub(Selectors, 'selectedRefinements').returns([
+        { navigationName: 'color', type: 'Value', value: 'blue' },
+        { navigationName: 'color', type: 'Value', value: 'red' }
+      ]);
+      stub(Selectors, 'config').returns(config);
+
+      const result = Adapter.convertBiasToSearch(state);
+
+      expect(result).to.deep.equal([{
```
effervescentia: just use `eql()`

test/unit/core/observer.ts
```diff
@@ -450,6 +451,24 @@ suite('Observer', ({ expect, spy, stub }) => {
           expect(emit).to.be.calledWith(Events.TEMPLATE_UPDATED, testObject);
         });
       });
+
+      describe('personalization', () => {
+        describe('_persist', () => {
+          describe('rehydrated', () => {
+            it('should emit PERSONALIZATION_BIASING_REHYDRATED if newState true', () => {
+              observers.data.present.personalization._persist.rehydrated(undefined, true);
```
effervescentia: does this need to be `undefined`? for the sake of the tests can we pass `null`?

test/unit/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,141 @@
...
+            }
+          },
+          allIds: [
+            { variant: 'a', key: 'z' },
+            ...state.biasing.allIds
+          ]
+        }
+      };
+      stub(personalization, 'insertSorted').returns(
+        newState.biasing.allIds
```
effervescentia: please make this one line

test/unit/core/selectors.ts
```diff
@@ -535,6 +536,113 @@ suite('selectors', ({ expect, stub }) => {
...
+    });
+
+    it('should be true on false/false', () => {
```
effervescentia: needs a more descriptive title

test/unit/core/store/middleware.ts
```diff
@@ -68,21 +76,14 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
       expect(applyMiddleware).to.be.calledWithExactly(
-        Middleware.thunkEvaluator,
-        validatorMiddleware,
-        idGeneratorMiddleware,
-        idGeneratorMiddleware,
-        errorHandlerMiddleware,
-        sagaMiddleware,
-        Middleware.thunkEvaluator,
-        Middleware.saveStateAnalyzer,
+        ...allMiddleware(),
```
effervescentia: can be one line

test/unit/core/store/middleware.ts
```diff
@@ -92,22 +93,14 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
-        Middleware.thunkEvaluator,
-        validatorMiddleware,
-        idGeneratorMiddleware,
-        idGeneratorMiddleware,
-        errorHandlerMiddleware,
-        sagaMiddleware,
-        Middleware.thunkEvaluator,
-        Middleware.saveStateAnalyzer
-      );
+        ...allMiddleware());
```
effervescentia: one line

test/unit/core/store/middleware.ts
```diff
@@ -218,4 +267,58 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+  describe('personalizationAnalyzer()', () => {
+    const conf = { a: 1 };
+    const state = { b: 2 };
+
+    it('should pass the action forward unchanged if not in list of relevant actions', () => {
+      const action = { a: 'b', type: 'NOT_VALID_ACTION' };
+      const config = stub(Selectors, 'config').returns(conf);
+      const enabled = stub(ConfigurationAdapter, 'isRealTimeBiasEnabled').returns(true);
+      const next = spy();
+      const getState = spy(() => state);
```
effervescentia: we made this a spy but aren't validating that it's been called

test/unit/core/store/middleware.ts
```diff
@@ -218,4 +267,58 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+      expect(config).to.be.calledWithExactly(state);
+      expect(enabled).to.be.calledWithExactly(conf);
+    });
+
+    it('should pass the action forward unchanged if real time biasing disabled', () => {
+      const action = { a: 'b', type: PERSONALIZATION_CHANGE_ACTIONS[0] };
+      const config = stub(Selectors, 'config').returns(conf);
+      const enabled = stub(ConfigurationAdapter, 'isRealTimeBiasEnabled').returns(false);
+      const next = spy();
+      const getState = spy(() => state);
```
effervescentia: spy not being validated

test/unit/core/store/middleware.ts
```diff
@@ -218,4 +267,58 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+
+    it('should make a batch action if action correct type', () => {
+      const action = { a: 'b', type: PERSONALIZATION_CHANGE_ACTIONS[0] };
+      const returnAction = 'return';
+      const extracted = 'extra';
+      const config = stub(Selectors, 'config').returns(conf);
+      const updateBiasing = stub(ActionCreators, 'updateBiasing').returns(returnAction);
+      const extract = stub(PersonalizationAdapter, 'extractBias').returns(extracted);
+      const next = spy();
+      const getState = spy(() => state);
```
effervescentia: spy not being validated

test/unit/core/store/middleware.ts
```diff
@@ -218,4 +267,58 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+      stub(ConfigurationAdapter, 'isRealTimeBiasEnabled').returns(true);
+
+      Middleware.personalizationAnalyzer(<any>{ getState })(next)(action);
+
+      expect(next).to.be.calledWith([
```
effervescentia: can be on line

src/core/adapters/personalization.ts
Filip-F: Yes, it gets used in the switch to decide what to do with the action.

src/core/adapters/personalization.ts
Filip-F: No, that specific byId should not be storing variant or key, as it is already indexed by that data.

src/core/sagas/products.ts
Filip-F: Don't we still need flux for once?

test/unit/core/store/middleware.ts
Filip-F: Fixed these 3.

test/unit/core/store/middleware.ts
Filip-F: Fixed

test/unit/core/store/middleware.ts
Filip-F: Fixed

src/core/actions/creators.ts
```diff
@@ -386,6 +393,21 @@ namespace ActionCreators {
...
+  /**
+   * The biasing object to receive and update biasing with
+   * @param  {Actions.Payload.Personalization.Biasing} payload - Biasing object
+   * @return {Actions.UpdateBiasing}
+   */
+  export function updateBiasing(payload: Actions.Payload.Personalization.Biasing) {
+    return (state: Store.State): Actions.UpdateBiasing =>
+      createAction(Actions.UPDATE_BIASING, {
+        ...payload,
+        config: Selectors.config(state).personalization.realTimeBiasing,
```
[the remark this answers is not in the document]
Filip-F: We need the config to know when we should prune data, as we're currently pruning both when we write new data and load from localstore. We didn't want to duplicate any data from the config into the store so we're getting this from the config.

src/core/adapters/personalization.ts
Filip-F: Fixed.

src/core/adapters/personalization.ts
Filip-F: We need to shallow clone the first n elements of allIds. Is there a way to do that with a spread?

src/core/adapters/personalization.ts
Filip-F: Don't want to modify the original array.

src/core/actions/creators.ts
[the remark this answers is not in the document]
effervescentia: If you see my note in the reducer we shouldn't need to pass through the config because the sorting will happen in the selector

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+  type ExtractableAction = Actions.SelectRefinement & Actions.AddRefinement;
+
+  export const DAYS_IN_SECONDS = 86400;
+
+  export const extractBias = (action: ExtractableAction, state: Store.State) => {
+    const config = Selectors.config(state).personalization.realTimeBiasing;
+    const byId = Selectors.realTimeBiasesById(state);
+    const { field, value } = extractRefinement(action, state);
+
+    if (!config.attributes[field]) {
```
effervescentia: turn this into a conditional return instead of a conditional early return

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+  export const extractRefinement = ({ type, payload }: ExtractableAction, state: Store.State): { field: string, value: string } => {
+    switch (type) {
+      case Actions.ADD_REFINEMENT:
+        if (!payload.range) {
+          return { field: payload.navigationId, value: payload.value };
+        }
+        break;
+      case Actions.SELECT_REFINEMENT:
+        return Selectors.refinementCrumb(state, payload.navigationId, payload.index);
+      default:
```
effervescentia: remove empty `default` case

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    return { field: undefined, value: undefined };
+  };
+
+  export const generateNewBias = () => ({
```
effervescentia: can be one line

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  // tslint:disable-next-line max-line-length
+  export const transformFromBrowser = (incomingState: BrowserStorageState, state: Store.State): Store.Personalization.Biasing => {
+    const config = Selectors.config(state);
+    // tslint:disable-next-line max-line-length
+    const olderThanTime = Math.floor(Date.now() / 1000) - DAYS_IN_SECONDS * ConfigAdapter.extractRealTimeBiasingExpiry(config);
+    const filteredIncomingState = incomingState.allIds.filter((element) => element.lastUsed >= olderThanTime);
+    let allIds = [];
+    const byId = {};
+    filteredIncomingState.forEach(({ variant, key, lastUsed }) => {
```
effervescentia: add a newline here to break up this function into setup, action and return

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+      allIds.push({ variant, key });
+      byId[variant] = { ...byId[variant], [key]: { lastUsed } };
+    });
+    Object.keys(byId).forEach((key) => {
+      allIds = pruneBiases(allIds, key, Object.keys(byId[key]).length, config.personalization.realTimeBiasing);
+    }); // filter allIds for each variant
+    return {
```
effervescentia: add newline here (also `return` can be one line)

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+      case Actions.ADD_REFINEMENT:
+        if (!payload.range) {
+          return { field: payload.navigationId, value: payload.value };
+        }
+        break;
+      case Actions.SELECT_REFINEMENT:
+        return Selectors.refinementCrumb(state, payload.navigationId, payload.index);
+      default:
+    }
+    return { field: undefined, value: undefined };
```
effervescentia: add newline before `return` to break up function

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+    const config = Selectors.config(state).personalization.realTimeBiasing;
+    const selectedRefinements = Selectors.selectedRefinements(state);
+
+    return allIds.filter(({ variant, key }) =>
+      !selectedRefinements.some(({ navigationName, type, value }) =>
+        type === 'Value' && navigationName === variant && key && value === key)
+    ).map(({ variant, key }) => ({
+      name: variant,
+      content: key,
+      strength: (config.attributes[variant] && config.attributes[variant].strength) || config.strength
```
effervescentia: ternary statement would be clearer

src/core/reducers/data/personalization.ts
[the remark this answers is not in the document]
effervescentia: rather than leave a `TODO` here, create a ticket in JIRA and remove this comment

src/core/sagas/products.ts
effervescentia: yea, but we should use the effects for `put()` as it makes it easier to test

src/core/selectors.ts
```diff
@@ -30,8 +30,8 @@ namespace Selectors {
    * Returns the corrected query if it exists, otherwise returns the original query.
    */
   export const currentQuery = (state: Store.State) => {
-    const query = state.data.present.query;
-    return query.corrected || query.original;
+    const stateQuery = state.data.present.query;
```
effervescentia:
```js
const { original, corrected } = /*..*/;
return corrected || original;
```

src/core/store/index.ts
```diff
@@ -338,6 +347,35 @@ namespace Store {
...
+
+    export interface BiasKey {
+      variant: string;
+      key: string;
+    }
+
+    export interface SingleBias {
+      // created: number;
+      lastUsed: number;
+      // timesClicked: number;
```
effervescentia: we can remove these comments for now

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,369 @@
...
+    });
+  });
+
+  describe('transformFromBrowser()', () => {
+    it('should filter allIds that are older than the set time', () => {
+      const oneDayInSec = 86400;
+      const now = Date.now() / 1000;
+      const expiry = 30;
+      const browserStorage = {
+        allIds: [{ variant: 'color', key: 'blue', lastUsed: now - oneDayInSec * 5 },
```
effervescentia: these arrays should be formatted like the rest of them:
```js
allIds: [
  { variant: 'color', /* ... */ },
  { /* ... */ }
]
```

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,369 @@
...
+                 { variant: 'brand', key: 'Shoe', lastUsed: now - oneDayInSec * 6 },
+                 { variant: 'color', key: 'red', lastUsed: now - oneDayInSec * 21 },
+                 { variant: 'brand', key: 'Nike', lastUsed: now - oneDayInSec * 33 }]
+      };
+      const biasFromBrowser = {
+        allIds: [{ variant: 'color', key: 'blue' }, { variant: 'brand', key: 'Shoe'},
```
effervescentia: array format here as well

test/unit/core/adapters/personalization.ts
```diff
@@ -0,0 +1,369 @@
...
+      const config = stub(Selectors, 'config').returns(conf);
+      const extractExpiry = stub(ConfigAdapter, 'extractRealTimeBiasingExpiry').returns(expiry);
+      const pruneBiases = stub(Adapter, 'pruneBiases').returnsArg(0);
+
+      const result = Adapter.transformFromBrowser(browserStorage, state);
+
+      expect(result).to.eql(biasFromBrowser);
+      expect(config).to.be.calledWithExactly(state);
+      expect(extractExpiry).to.be.calledWithExactly(conf);
+      expect(pruneBiases).to.be.calledTwice.and
```
effervescentia: move `.and`s to precede the additional expectations

test/unit/core/sagas/products.ts
```diff
@@ -413,6 +415,40 @@ suite('products saga', ({ expect, spy, stub }) => {
...
+        expect(getState).to.be.calledWith();
+        expect(dispatch).to.be.calledWith(payload);
+      });
+
+      it('should wait on personalization biasing rehydrated if data not loaded from browser', () => {
+        const once = spy();
+        const dispatch = spy();
+        const getState = spy();
+        stub(Selectors, 'realTimeBiasesHydrated').returns(false);
+
```
effervescentia: can this be moved up to the rest of the `const`s

test/unit/core/selectors.ts
effervescentia: ^^^

test/unit/core/store/index.ts
```diff
@@ -36,6 +38,29 @@ suite('Store', ({ expect, spy, stub }) => {
         .and.calledWith('m')
         .and.calledWith('n')
         .and.calledWith('o');
+      // expect(persistStore).to.be.called;
```
effervescentia: no need to leave this in if we can't stub it

test/unit/core/store/index.ts
```diff
@@ -36,6 +38,29 @@ suite('Store', ({ expect, spy, stub }) => {
...
+    });
+
+    it('should not persist store if real time bias disabled', () => {
+      // tslint:disable-next-line variable-name
+      const __config = { a: 'b' };
```
effervescentia: this `const` is not re-used, so it can just provide its value inline in `flux`

test/unit/core/store/index.ts
```diff
@@ -36,6 +38,29 @@ suite('Store', ({ expect, spy, stub }) => {
...
+      const flux: any = { __config };
+      const state = { c: 'd' };
+      const middleware = ['e', 'f', 'g'];
```
effervescentia: do we need to save this `const` for later in the test?

test/unit/core/store/index.ts
```diff
@@ -36,6 +38,29 @@ suite('Store', ({ expect, spy, stub }) => {
...
+      const storeInstance = { h: 'i' };
```
effervescentia: look like we don't re-use this `const` either

test/unit/core/store/middleware.ts
```diff
@@ -167,14 +154,70 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
       Middleware.errorHandler(<any>{})(null)(next)(action);
 
       expect(next).to.be.calledWith(undoAction);
     });
   });
 
+  describe('saveStateAnalyzer()', () => {
+    it('should pass through other actions', () => {
+      const next = spy();
+      const getState = spy();
```
effervescentia: we aren't making any expectations about `next()` or `getState()` so why do they need to be `spy()`s? and why do we need to save them as `const`s at all?

test/unit/core/store/middleware.ts
effervescentia: 👍

test/unit/core/store/middleware.ts
```diff
@@ -218,4 +261,78 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+      const enabled = stub(ConfigurationAdapter, 'isRealTimeBiasEnabled').returns(true);
+      const next = spy();
+      const getState = spy(() => state);
+
+      Middleware.personalizationAnalyzer(<any>{ getState })(next)(action);
+
+      expect(next).to.be.calledWithExactly(action);
+      expect(config).to.be.calledWithExactly(state);
+      expect(enabled).to.be.calledWithExactly(conf);
+      expect(getState).to.be.calledWithExactly();
```
effervescentia: can just expect to be `called` if we don't have any params to validate

test/unit/core/store/middleware.ts
```diff
@@ -218,4 +261,78 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+      const next = spy();
+      const getState = spy(() => state);
+      stub(ConfigurationAdapter, 'isRealTimeBiasEnabled').returns(true);
+
+      Middleware.personalizationAnalyzer(<any>{ getState })(next)(action);
+
+      expect(next).to.be.calledWithExactly(action);
+      expect(extract).to.be.calledWithExactly(action, state);
+      expect(updateBiasing).to.not.be.called;
+      expect(config).to.be.calledWithExactly(state);
```
effervescentia: we also don't need to duplicate all of these tests out for every `it()` as we've already covered them in other tests

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+  // tslint:disable-next-line max-line-length
+  export const transformFromBrowser = (incomingState: BrowserStorageState, state: Store.State): Store.Personalization.Biasing => {
+    const config = Selectors.config(state);
+    // tslint:disable-next-line max-line-length
+    const olderThanTime = Math.floor(Date.now() / 1000) - DAYS_IN_SECONDS * ConfigAdapter.extractRealTimeBiasingExpiry(config);
+    const filteredIncomingState = incomingState.allIds.filter((element) => element.lastUsed >= olderThanTime);
+    let allIds = [];
+    const byId = {};
+    filteredIncomingState.forEach(({ variant, key, lastUsed }) => {
+      allIds.push({ variant, key });
```
effervescentia: rather than pushing everything into `allIds` then pruning them back out, I'd prefer if you only conditionally add them

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
+import Actions from '../actions';
+import Configuration from '../configuration';
+import Selectors from '../selectors';
+import Store from '../store';
+import ConfigAdapter from './configuration';
+
+namespace Personalization {
```
effervescentia: In general in this file we have a bunch of references to `variant` which is a very specific and different concept from what we want to be referring to here which is `refinements`

everywhere that we have `key` and `variant` I would prefer the wording `field` and `value` which is what we use everywhere else

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+    // tslint:disable-next-line max-line-length
+    const olderThanTime = Math.floor(Date.now() / 1000) - DAYS_IN_SECONDS * ConfigAdapter.extractRealTimeBiasingExpiry(config);
+    const filteredIncomingState = incomingState.allIds.filter((element) => element.lastUsed >= olderThanTime);
+    let allIds = [];
+    const byId = {};
+    filteredIncomingState.forEach(({ variant, key, lastUsed }) => {
+      allIds.push({ variant, key });
+      byId[variant] = { ...byId[variant], [key]: { lastUsed } };
+    });
+    Object.keys(byId).forEach((key) => {
```
effervescentia: shouldn't need this if conditionally adding rather than pruning after the fact

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,117 @@
...
+    }))
+  });
+
+  // tslint:disable-next-line max-line-length
+  export const transformFromBrowser = (incomingState: BrowserStorageState, state: Store.State): Store.Personalization.Biasing => {
+    const config = Selectors.config(state);
+    // tslint:disable-next-line max-line-length
+    const olderThanTime = Math.floor(Date.now() / 1000) - DAYS_IN_SECONDS * ConfigAdapter.extractRealTimeBiasingExpiry(config);
+    const filteredIncomingState = incomingState.allIds.filter((element) => element.lastUsed >= olderThanTime);
+    let allIds = [];
```
effervescentia: should be able to be an array when pruning removed

src/core/reducers/data/personalization.ts
```diff
@@ -0,0 +1,57 @@
...
+export const updateBiasing = (state: State, payload: Actions.Payload.Personalization.Biasing) => {
+  const byId = {
+    ...state.biasing.byId,
+    [payload.variant]: {
+      ...state.biasing.byId[payload.variant],
+      [payload.key]: payload.bias
+    }
+  };
+
+  const allIds = Adapter.pruneBiases(insertSorted(state.biasing.allIds, { variant: payload.variant, key: payload.key }),
```
effervescentia: not sure if the comment got removed somehow but I don't want to do the sorting before we put the data in the store, only when selecting it out of the store, this also means we don't need to pass the `config` with this action

test/unit/core/store/middleware.ts
Filip-F: Done.

test/unit/core/store/middleware.ts
Filip-F: Next is being used.

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    return { field: undefined, value: undefined };
+  };
+
+  export const generateNewBias = () => ({ lastUsed: Math.floor(Date.now() / 1000) });
+
+  // tslint:disable-next-line max-line-length
+  export const transformToBrowser = (state: Store.Personalization.Biasing, reducerKey: string): BrowserStorageState => ({
+    allIds: state.allIds.map(({ variant, key }) => ({
+      variant,
```
effervescentia: I don't want to confuse the usage of the term `variant` which is specifically for a permutation of a product (perhaps a version of a product that comes in a given colour). The concept we want to use here is refinements which we represent in the rest of the system as `field` and `value` which I would like to use instead of `key` and `variant`

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,120 @@
...
+      name: variant,
+      content: key,
+      strength: config.attributes[variant].strength || config.strength,
+    }));
+  };
+
+  // tslint:disable-next-line max-line-length
+  export const pruneBiases = (allIds: Store.Personalization.BiasKey[], variant: string, variantCount: number, config: Configuration.Personalization.RealTimeBiasing) => {
+    if (config.attributes[variant] && variantCount >= config.attributes[variant].maxBiases) {
+      for (let i = allIds.length - 1; i >= 0; i--) {
```
effervescentia: this could be simplified to
```js
const index = allIds.reverse().findIndex((id) => id.variant === variant);
if (index !== -1) {
  return [...allIds.slice(0, index), ...allIds.slice(index + 1)];
}
```

src/core/adapters/personalization.ts
```diff
@@ -0,0 +1,120 @@
...
+        type === 'Value' && navigationName === variant && key && value === key)
+    ).map(({ variant, key }) => ({
+      name: variant,
+      content: key,
+      strength: config.attributes[variant].strength || config.strength,
+    }));
+  };
+
+  // tslint:disable-next-line max-line-length
+  export const pruneBiases = (allIds: Store.Personalization.BiasKey[], variant: string, variantCount: number, config: Configuration.Personalization.RealTimeBiasing) => {
```
effervescentia: I would rename `allIds` to a more appropriate name such as `biasKeys`

src/core/adapters/personalization.ts
Filip-F: That mutates allIds though, which we can't do since we're using this on reducers. What we could do is declare a new variable like `const ids = allIds.slice();` at the start then mutate that, if you think that sounds alright.

src/core/adapters/personalization.ts
Filip-F: Fixed.

PR #127, opened 2017-11-08
fix(BBD 811): Search is broken or haunted

Filip-F:
Related:
* https://github.com/groupby/flux-capacitor/pull/127
* https://github.com/groupby/storefront-core/pull/139
* https://github.com/groupby/storefront-details/pull/18
* https://github.com/groupby/storefront-products/pull/26

### DIFF
--- a/src/core/reducers/data/details.ts
+++ b/src/core/reducers/data/details.ts
@@ -1,7 +1,7 @@
 import Actions from '../../actions';
 import Store from '../../store';
 
-export type Action = Actions.UpdateDetails | Actions.ReceiveDetailsProduct;
+export type Action = Actions.UpdateDetails;
 export type State = Store.Details;
 
 export const DEFAULTS: State = {};
@@ -9,13 +9,9 @@ export const DEFAULTS: State = {};
 export default function updateDetails(state: State = DEFAULTS, action: Action): State {
   switch (action.type) {
     case Actions.UPDATE_DETAILS: return update(state, action.payload);
-    case Actions.RECEIVE_DETAILS_PRODUCT: return receiveProduct(state, action.payload);
     default: return state;
   }
 }
 
 export const update = (state: State, data: Store.Product) =>
   ({ ...state, data });
-
-export const receiveProduct = (state: State, product: Store.Product) =>
-  ({ ...state, product });
--- a/src/core/sagas/product-details.ts
+++ b/src/core/sagas/product-details.ts
@@ -32,11 +32,10 @@ export namespace Tasks {
     }
   }
 
-  export function* receiveDetailsProduct(flux: FluxCapacitor, { payload: record }: Actions.SetDetails) {
-    let product = record;
-    if (record.allMeta) {
-      flux.emit(Events.BEACON_VIEW_PRODUCT, record);
-      product = record.allMeta;
+  export function* receiveDetailsProduct(flux: FluxCapacitor, { payload: product }: Actions.SetDetails) {
+    if (product.allMeta) {
+      flux.emit(Events.BEACON_VIEW_PRODUCT, product);
+      product = product.allMeta;
     }
     yield effects.put(flux.actions.updateDetails(product));
     flux.saveState(utils.Routes.DETAILS);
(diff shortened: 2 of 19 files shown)


test/unit/core/selectors.ts
```diff
@@ -277,6 +277,15 @@ suite('selectors', ({ expect, stub }) => {
     });
   });
 
+  describe('productWithId()', () => {
+    it('should find product with given id', () => {
+      const products = [{ id: '3' }, { id: '4' }, { id: '7' }, { id: '2' }];
+      const productSelector = stub(Selectors, 'products').returns(products);
```
zfcb: Is this const needed?

Filip-F: Forgot to add an expect.

src/core/sagas/product-details.ts
```diff
@@ -21,25 +21,29 @@ export namespace Tasks {
...
-  export function* receiveDetailsProduct(flux: FluxCapacitor, record: Store.Product) {
-    flux.emit(Events.BEACON_VIEW_PRODUCT, record);
-    yield effects.put(flux.actions.receiveDetailsProduct(record.allMeta));
+  export function* receiveDetailsProduct(flux: FluxCapacitor, { payload: record }: Actions.SetDetails) {
+    let product = record;
+    if (record.allMeta) {
+      flux.emit(Events.BEACON_VIEW_PRODUCT, record);
+      product = record.allMeta;
+    }
+    yield effects.put(flux.actions.updateDetails(product));
```
zfcb: How about record = record.allmeta on line 39 and maybe change { payload: record } to { payload: product }

zfcb (dismissed):

src/flux-capacitor.ts
```diff
@@ -98,6 +98,10 @@ class FluxCapacitor extends EventEmitter {
     this.store.dispatch(this.actions.updateDetails(product));
```
Vic-Dev: Is this action still used anywhere?

Vic-Dev: Because if it isn't, you could just make the new action replace this one.

Vic-Dev (changes requested):
Looks good. Some comments.

src/core/reducers/data/details.ts
```diff
@@ -1,21 +1,17 @@
...
     case Actions.UPDATE_DETAILS: return update(state, action.payload);
-    case Actions.RECEIVE_DETAILS_PRODUCT: return receiveProduct(state, action.payload);
     default: return state;
   }
 }
 
 export const update = (state: State, data: Store.Product) =>
   ({ ...state, data });
-
-export const receiveProduct = (state: State, product: Store.Product) =>
```
Vic-Dev: Because you're removing this section, I think there's actually some cleanup you need to do surrounding the products section of details in the store, because I don't think it's ever used now. This means you also need to clean up UPDATE_PRODUCT_DETAILS action.

Vic-Dev (changes requested):
One more important thing

src/core/reducers/data/details.ts
Filip-F: Because gb-product expects an untransformed product so it can run productTransformer on it. If given an already transformed product it gives this: https://i.imgur.com/DCKnWJs.png

Filip-F: Did you mean FETCH_PRODUCT_DETAILS? I already removed UPDATE_PRODUCT_DETAILS. The former is still needed, however, as when first loading a page you need to call that to get the full product dataset using the data in the url. It could be renamed to FETCH_DETAILS but I wanted to to avoid breaking things for clients. I've removed every superfluous action for product details however.

Vic-Dev: mb wrong name `DETAILS_PRODUCT_UPDATED`

Vic-Dev (dismissed):
That's a 🍇  job!

Vic-Dev:
Can you demo it? can't remember if you ahve

Vic-Dev (approved):
👂 -ily good stuff!

PR #139, opened 2017-12-06
feat(BBD-833): Connect CVS Past Purchase work to live API

Filip-F:
https://github.com/groupby/flux-capacitor/pull/139
https://github.com/groupby/storefront-core/pull/150
https://github.com/groupby/storefront-products/pull/28
https://github.com/groupby/storefront-sort/pull/18
https://github.com/groupby/storefront-structure/pull/41
https://github.com/groupby/storefront-navigation/pull/32
https://github.com/groupby/storefront-recommendations/pull/8
https://github.com/groupby/storefront-presets/pull/52


src/core/sagas/recommendations.ts
```diff
@@ -49,56 +59,110 @@ export namespace Tasks {
...
+    if (securedPayload) {
+      const url = `https://${config.customerId}.groupbycloud.com/orders/v1/public/skus/${endpoint}`;
+      const response = yield effects.call(utils.fetch, url, Adapter.buildBody({
+        securedPayload,
+        query
+      }));
+      return yield response.json();
+    }
+    throw new MissingPayload();
+  }
```
treeforever: why is creating new error message so complicated? Can't we do just throw new Error('missing payload')?

src/core/sagas/recommendations.ts
```diff
@@ -49,56 +59,110 @@ export namespace Tasks {
...
+        yield effects.put(flux.actions.fetchPastPurchaseNavigations());
+      } else {
+        yield effects.put(flux.actions.receivePastPurchaseSkus([]));
       }
-      return [];
     } catch (e) {
-      return effects.put(flux.actions.receivePastPurchases(e));
+      if (!(e instanceof MissingPayload)) { // pass through misisng payloads
+        return effects.put(flux.actions.receivePastPurchaseSkus(e));
+      }
```
treeforever: what happens if it is an instance of MissingPayload?

src/core/sagas/recommendations.ts
```diff
@@ -49,56 +59,110 @@ export namespace Tasks {
...
+      const pastPurchaseSkus = yield effects.select(Selectors.pastPurchases);
+      if (pastPurchaseSkus.length > 0) {
+        const request = yield effects.select(Requests.autocompleteProducts);
+        const results = yield effects.call(fetchProductsFromSkus, flux, pastPurchaseSkus, {
+          ...request,
+          query: payload
+        });
+        yield effects.put(flux.actions.receiveSaytPastPurchases(SearchAdapter.augmentProducts(results)));
+      } else {
+        yield effects.put(flux.actions.receiveSaytPastPurchases([]));
```
treeforever: so when pastPurchaseSkus is empty, flux still could issue the action FETCH_SAYT_PAST_PURCHASES? Can you eliminate this possibility from the beginning so you don't need to deal with it here?

src/core/adapters/configuration.ts
```diff
@@ -13,27 +15,26 @@ namespace Adapter {
...
-              values: []
             },
           },
-          recommendations: {
-            suggested: { products: [] },
-            pastPurchases: { products: []},
-            queryPastPurchases: [],
-            orderHistory: []
-          },
           fields: Adapter.extractFields(config),
```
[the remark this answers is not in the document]
zfcb: This is only needed here when a config property is required, otherwise the initial state is obtained from the reducers.

src/core/sagas/recommendations.ts
Filip-F: It gets ignored. I'm passing through missing payload errors and simply allowing the action to do nothing since there's no need to roll back the state.

src/core/sagas/recommendations.ts
zfcb: I think this may be the cleanest way, rather than have that logic at every instance of FETCH_SAYT_PAST_PURCHASES being dispatched.

treeforever: but why would FETCH_SAYT_PAST_PURCHASES being dispatched when pastPurchaseSkus is empty? Is it when skus are cleared out? Just trying to understand here.

src/core/actions/index.ts
```diff
@@ -32,6 +33,9 @@ namespace Actions {
     (...args: any[]): Action | Action[] | Thunk<any>;
   }
 
+  export import Payload = Payloads;
```
effervescentia: did not know you could do this, very interesting

src/core/adapters/recommendations.ts
```diff
@@ -108,12 +109,51 @@ namespace Recommendations {
...
+
+  export const pastPurchaseNavigations = (config: Configuration, navigations: Store.Navigation[]) => {
+    const configNavigations = ConfigurationAdapter.extractPastPurchaseNavigations(config);
+
+    return navigations.reduce((acc, navigation) => {
+      return configNavigations[navigation.field] ?
+        acc.concat({
+          ...navigation,
+          refinements: configNavigations[navigation.field].length > 0 ?
+            navigation.refinements.reduce((refinementAcc, refinement: Store.ValueRefinement) => {
```
effervescentia: very dense logic here, would prefer if this could be modularized

src/core/sagas/recommendations.ts
```diff
@@ -1,13 +1,23 @@
...
 import Adapter from '../adapters/recommendations';
 import SearchAdapter from '../adapters/search';
 import Configuration from '../configuration';
 import Requests from '../requests';
 import Selectors from '../selectors';
 import Store from '../store';
-import { fetch } from '../utils';
+import * as utils from '../utils';
+
+export class MissingPayload extends Error {
```
effervescentia: can we name this `MissingPayloadError`

PR #141, opened 2017-12-12
Greenkeeper/@types/sinon 4.1.1

jatkinson10:


test/unit/flux-capacitor.ts
```diff
@@ -83,7 +83,7 @@ suite('FluxCapacitor', ({ expect, spy, stub }) => {
     let actions: typeof FluxCapacitor.prototype.actions;
     let store: any;
 
-    function expectDispatch(action: () => void, creatorName: string, ...params: any[]) {
+    function expectDispatch(action: () => void, creatorName: any, ...params: any[]) {
```
Filip-F: Why was this changed from string to any?

PR #144, opened 2017-12-12
[code review only] feat(BBD-825): save cart status in store

treeforever:
currently does:
- save tracker info in cart state
- create cart and save cart id in state
- add to cart
- use persist to save to browser

doesn't do: 
- transform when writing to and reading from browser
- API calls (code in saga is stubbed)
- consolidate data in the cart

### DIFF
--- a/src/core/adapters/cart.ts
+++ b/src/core/adapters/cart.ts
@@ -16,15 +16,19 @@ namespace Cart {
     body: JSON.stringify(body)
   });
 
-  export const transformToBrowser = (state: Store.Cart): any => ({
-    cartId: state.cartId
-  });
-}
+  export const transformToBrowser = (state: Store.Cart): any => state;
 
-export interface CartBody {
-  loginId?: string;
-  sessionId: string;
-  visitorId: string;
+  // tslint:disable-next-line:max-line-length
+  export const transformFromBrowser = (incomingState: Store.Cart): Store.Cart => {
+    return incomingState;
+  };
+
+  export interface CartBody {
+    loginId?: string;
+    sessionId: string;
+    visitorId: string;
+  }
+
 }
 
 export default Cart;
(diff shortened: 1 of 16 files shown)


src/core/configuration.ts
```diff
@@ -221,7 +221,7 @@ namespace Configuration {
     location?: Configuration.Recommendations.Location;
 
     /**
-     * Product ID field as used in recommendations.
+     * Product ID field as used in recommendations.`
```
zfcb: Quasiquote added here?

src/core/observer.ts
```diff
@@ -147,6 +147,13 @@ namespace Observer {
             orderHistory: emit(Events.ORDER_HISTORY_UPDATED)
           },
 
+          cart: {
+            content: {
+              cartId: emit(Events.CART_ID_UPDATED),
+              quantity: emit(Events.CART_UPDATED)
```
zfcb: Rename event to cart quantity updated

src/core/adapters/cart.ts
```diff
@@ -0,0 +1,39 @@
...
+  export const transformToBrowser = (state: Store.Cart, reducerKey: string): any => state;
+
+  // tslint:disable-next-line:max-line-length
+  export const transformFromBrowser = (incomingState: Store.Cart): Store.Cart => {
+    return incomingState;
+  };
+
+  export const calculateQuantity = (items: any[]) =>
+    items.reduce((acc, item) => {
+      return acc + item['quantity'];
```
zfcb: I think calculateTotalQuantity would be a better name

src/core/adapters/configuration.ts
```diff
@@ -33,7 +33,7 @@ namespace Adapter {
           page: {
             ...PageReducer.DEFAULTS,
             sizes: Adapter.extractPageSizes(config, PageReducer.DEFAULT_PAGE_SIZE)
-          },
+          }
```
zfcb: Comma removed here?

treeforever: why not?

PR #153, opened 2018-01-11
feat(BBD-825): shopping cart

treeforever:


src/core/actions/index.ts
```diff
@@ -206,6 +207,20 @@ namespace Actions {
...
+  export const CART_CREATED = 'CART_CREATED';
+  export type CartCreated = Action<typeof CART_CREATED, string>;
+  export const ADD_TO_CART = 'ADD_TO_CART';
+  export type AddToCart = Action<typeof ADD_TO_CART, Store.CartProduct>;
+  export const CART_SERVER_UPDATED = 'CART_SERVER_UPDATED';
+  export type CartServerUpdated = Action<typeof CART_SERVER_UPDATED, Partial<Store.Cart>>;
+  export const ITEM_QUANTITY_CHANGED = 'ITEM_QUANTITY_CHANGED';
+  export type ItemQuantityChanged = Action<typeof ITEM_QUANTITY_CHANGED, any>;
+  export const REMOVE_ITEM = 'REMOVE_ITEM';
+  export type RemoveItem = Action<typeof REMOVE_ITEM, Store.CartProduct>;
```
Vic-Dev: Should maybe include the word "cart" with all the actions, like we do with other action types. The reason for this, is what if in the future there's a `Wishlist` api, which would be very similar to cart (needing the same actions, etc.) so we need a way of differentiating that this is the CartRemoveItem action, etc.

test/unit/core/adapters/cart.ts
```diff
@@ -0,0 +1,218 @@
...
+      },
+      {
+        sku: '456',
+        quantity: 2,
+        title: 'fan'
+      }];
+      const item = <any>{
+        sku: 'unique',
+        quantity: 2,
+        title: 'unique cat'
```
Vic-Dev: Gotta watch out for those unique cats lol

test/unit/core/adapters/cart.ts
```diff
@@ -0,0 +1,218 @@
...
+      }];
+
+      const result = Adapter.combineLikeItems(stateItems, item);
+
+      expect(result).to.eql(expected);
+    });
+  });
+
+  describe('calculateTotalQuantity()', () => {
```
Vic-Dev: Move to the top

test/unit/core/reducers/data/cart.ts
```diff
@@ -0,0 +1,161 @@
...
+      sessionId: '',
+      generatedTotalPrice: 0,
+      lastModified: null,
+    }
+  };
+
+  describe('updateCart()', () => {
+    it('should update cart state on GET_TRACKER_INFO', () => {
+      const payload = { visitorId: 'v', sessionId: 's' };
+      const newState = {
```
Vic-Dev: You could just have made `newState = { ...state, visitorId: 'v', sessionId: 's' }`
No need to change it now, since you've written all of this, but in the future, just a heads up it can be simplified.

test/unit/core/reducers/data/cart.ts
```diff
@@ -0,0 +1,161 @@
...
+          generatedTotalPrice: 100,
+          lastModified: 123456,
+        }
+      };
+
+      const reducer = cart(state, { type: Actions.CART_SERVER_UPDATED, payload: content });
+
+      expect(reducer).to.eql(newState);
+    });
+
```
Vic-Dev: Add test for `ITEM_QUANTITY_CHANGED`

Vic-Dev (changes requested):
Great job again! :) 🔫 -na miss u mengqi

jmbtutor:
Archiving repo. Closing.

PR #154, opened 2018-01-12
fix(BBD-901): Selected navigation and refinement should appear at top of list rather than bottom

Vic-Dev:


src/core/adapters/search.ts
```diff
@@ -101,9 +101,10 @@ namespace Adapter {
         const navigation = Adapter.extractNavigation(selectedNav);
         navigation.selected = <any[]>Object.keys(Array(navigation.refinements.length).fill(0))
           .map((value) => Number(value));
-        navigations[selectedNav.name] = navigation;
+        navigations = { [selectedNav.name]: navigation, ...navigations };
```
treeforever: I tried this code but saw one navigation not necessarily go to the top, or when another navigation is selected, other selected navigations don't stay at where they are. So it seems the order is a little unpredictable based on this solution.

I think the reason is the order of navigations is decided by the order of the navigations.allIds array in the state, not necessarily by the order of object (object keys don't have a order). So it seems a little complicated. Maybe map navigations to an array so it will have a fixed order?

Vic-Dev: Yes, I agree that the problem is using Object.keys, but I didn't see this inconsistent behaviour. We should demo/you can show me what you mean. But yes, I agree there is a better solution, with a lot of refactoring.

PR #155, opened 2018-01-19
feat(BBD-543): Infinite-scroll

Vic-Dev:
Switch to using currentRecordCount w/ pastpurchase rather than allRecordCount


src/core/reducers/ui.ts
```diff
@@ -18,12 +18,16 @@ export default function updateUi(state: State = {}, { type, payload, meta = <Act
...
 
-export const createComponentState = (state: State, { tagName, id, state: tagState }: Actions.Payload.Component.State) =>
+// tslint:disable-next-line max-line-length
+export const createComponentState = (state: State, { tagName, id, state: tagState, persist }: Actions.Payload.Component.State) =>
   ({
     ...state,
     [tagName]: {
       ...state[tagName],
-      [id]: tagState
+      [id]: {
```
treeforever: why the change?

src/core/requests.ts
```diff
@@ -39,13 +39,14 @@ namespace Requests {
     return <Request>Requests.chain(config.search.defaults, request, config.search.overrides);
   };
 
-  export const pastPurchaseProducts = (state: Store.State, getNavigations: boolean = false): Request => {
+  // tslint:disable-next-line max-line-length
+  export const pastPurchaseProducts = (state: Store.State, getNavigations: boolean = false, options: { pageSize: number, skip: number} = { pageSize: Selectors.pastPurchasePageSize(state), skip: Selectors.pastPurchasePageSize(state) * (Selectors.pastPurchasePage(state) - 1) }): Request => {
```
treeforever: Consider to pass in this long list of parameters as an object or get default value in the body of the function. It causes difficulty to read when it gets long.

treeforever: One question: when you pass in options, what if `pageSize` is defined and `skip` is undefined, it won't get the default value I assume? Or maybe this scenario won't occur?

treeforever:
I also have a general question. Since we've seen some clients with massive data for products blowing up firefox and IE, it may be worth considering how infinite scroll can hold so much products data in the state.

src/core/reducers/ui.ts
Vic-Dev: Wanted to add a "persist" flag so that info could persist over time

src/core/requests.ts
Vic-Dev: You're right, it won't work. But! If you use this syntax, it does:
`{ pageSize = Selectors.pastPurchasePageSize(state), skip = Selectors.pastPurchasePageSize(state) * (Selectors.pastPurchasePage(state) - 1) }: { pageSize?: number, skip?: number}`

I switched it to not passing it in as default though, because you're right that line's getting too long. :) Just thought that syntax was kewl.

src/core/store/index.ts
```diff
@@ -90,7 +92,10 @@ namespace Store {
   export interface UI {
     [tagName: string]: {
       global?: any;
-      [tagId: string]: any;
+      [tagId: number]: {
```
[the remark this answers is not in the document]
Vic-Dev: This was going to be for infinite isFetchingForward & isFetchingBackward state, but I ended up creating a section in the store for InfiniteScroll.

I can take it out if you think I should.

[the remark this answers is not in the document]
treeforever: Ok I see. If it is not used then we should take it out I think :)

PR #156, opened 2018-01-19
fix(BBD-898): Storefront url redirect service not working for Firefox and IE11

treeforever:
https://github.com/groupby/storefront-core/pull/163


src/core/reducers/index.ts
```diff
@@ -1,6 +1,7 @@
 import * as redux from 'redux';
-import undoable from 'redux-undo';
+import { default as undoable, includeAction } from 'redux-undo';
```
Vic-Dev: I think the style we use in other places is `import undoable, { includeAction } from 'redux-undo';`

src/core/reducers/index.ts
```diff
@@ -9,15 +10,32 @@ import ui from './ui';
...
+    const limit: number = ConfigAdapter.extractHistoryLength(store);
+    const config = {
+      limit,
+      filter: includeAction(Actions.SAVE_STATE),
+    };
+    const reducer = undoable(data, config);
+    const { history, ...newState } = reducer(state, action);
+
+    if (limit === 0) {
+      return { ...newState, past: [{}] };
```
Vic-Dev: why pass an empty object to past?

treeforever: oh I probably should leave a comment there. Because redux-undo takes 0 (limit) as no limit. So if the history length coming from cofig is 0, then I manually overwrite past as an empty array. What's more, I added {} in the array because it is required so in one of the tests, so I figured it's best to keep the same format.

src/core/reducers/index.ts
treeforever: I need to import 'redux-undo'.default as undoable, therefore the current style. Maybe there's a better way of writing it?

Vic-Dev: That's what my syntax is doing

src/core/reducers/index.ts
Vic-Dev: 👍

PR #164, opened 2018-01-25
fix(BBD-927): fix overrides does not seem to apply to products refinements through SAYT

treeforever:


src/core/sagas/autocomplete.ts
```diff
@@ -66,12 +67,16 @@ export namespace Tasks {
   export function* fetchProducts(flux: FluxCapacitor, { payload: { query, refinements } }: Actions.FetchAutocompleteProducts) {
     try {
       const request = yield effects.select(Requests.autocompleteProducts);
+      const overrideRefinements = request.refinements;
+      const originalRefinements = refinements.map(({ field, value }) =>
+        ({ type: 'Value', navigationName: field, value }));
```
vescogma: This seems like it might miss the exclude flag. I mean I know we currently don't handle it, but wouldn't we want to future proof this?

PR #171, opened 2018-02-06
firing a generic event alongside the specific one

interface-master:


test/unit/core/observer.ts
```diff
@@ -341,6 +341,7 @@ suite('Observer', ({ expect, spy, stub }) => {
           observers.data.present.navigations(oldNavigations, newNavigations);
 
           expect(emit).to.be.calledWith(`${Events.SELECTED_REFINEMENTS_UPDATED}:c`, newNavigations.byId.c);
+          expect(emit).to.be.calledWith(`${Events.SELECTED_REFINEMENTS_UPDATED}`, newNavigations.byId.c);
```
effervescentia: don't need a template string here

test/unit/core/observer.ts
```diff
@@ -356,6 +357,7 @@ suite('Observer', ({ expect, spy, stub }) => {
...
```
effervescentia: don't need a template string here

src/core/observer.ts
```diff
@@ -66,6 +66,7 @@ namespace Observer {
                     || oldNavigation.refinements !== newNavigation.refinements) {
                     // tslint:disable-next-line max-line-length
                     emit(`${selectedRefinementsEvent}:${id}`)(oldNavigation, newNavigation, `${path}.byId.${id}`);
+                    emit(`${selectedRefinementsEvent}`)(oldNavigation, newNavigation, `${path}.byId.${id}`);
```
effervescentia: don't need a template string here
should just pass `path` as third param

PR #175, opened 2018-05-14
EN-410 - added support for ProxyBridge

radichc:


src/flux-capacitor.ts
```diff
@@ -161,7 +161,11 @@ class FluxCapacitor extends EventEmitter {
    */
   static createBridge(config: Configuration, errorHandler: (err: Error) => void) {
     const networkConfig = config.network;
-    const bridge = new BrowserBridge(config.customerId, networkConfig.https, networkConfig);
+
+    const bridge = networkConfig.proxyUrl ?
```
effervescentia: should just be `const bridge = new BrowserBridge(config.customerId, networkConfig.https, networkConfig);` once the other change has been made

PR #179, opened 2018-07-20
fix(SF-1013): Be more defensive when loading more products

jmbtutor:


src/core/sagas/products.ts
```diff
@@ -100,12 +100,19 @@ export namespace Tasks {
       const state: Store.State = yield effects.select();
       const products = Selectors.productsWithMetadata(state);
       const pageSize = action.payload.amount;
+      const recordCount = Selectors.recordCount(state);
 
       let skip;
       if (action.payload.forward) {
-        skip = products[products.length - 1].index;
+        skip = products.length === 0 ? 0 : products[products.length - 1].index;
```
jmbtutor: Alternatively, `products.length && products[products.length - 1].index`, but I don't think it's as clear what the intent is.

src/core/reducers/data/products.ts
```diff
@@ -14,7 +14,11 @@ export default function updateProducts(state: State = [], action: Action): State
 
 // tslint:disable-next-line max-line-length
 export function updateMoreProducts(state: State, { payload }: Actions.ReceiveMoreProducts | Actions.ReceiveMorePastPurchaseProducts) {
-  if (state[0].index > payload[payload.length - 1].index) {
+  if (payload.length === 0) {
```
Vic-Dev: Just wondering why you chose to be so explicit (rather than just `payload.length`)

src/core/reducers/data/products.ts
```diff
@@ -14,7 +14,11 @@ export default function updateProducts(state: State = [], action: Action): State
...
+    return state;
+  } else if (state.length === 0) {
```
Vic-Dev: Same here

PR #180, opened 2018-07-22
feat(SF-1011): Add realTimeBiasing to autocomplete products

Vic-Dev:


src/core/requests.ts
```diff
@@ -68,15 +69,33 @@ namespace Requests {
...
+  };
+
+  export const realTimeBiasing = (state: Store.State, request: Request): Request => {
+    const addedBiases = Personalization.convertBiasToSearch(state);
+
+    return {
+      ...request,
+      biasing: {
+        ...request.biasing,
+        biases: ((request.biasing ? request.biasing.biases : []) || []).concat(addedBiases),
```
jmbtutor: `(request.biasing && request.biasing.biases || []).concat(addedBiases)`, although I see that it was moved from `products.ts`, so up to you if you want to change it.

test/unit/core/sagas/products.ts
```diff
@@ -463,71 +463,19 @@ suite('products saga', ({ sinon, expect, spy, stub }) => {
         const receiveProductsAction: any = { c: 'd' };
         const biases = ['bias'];
         const request = { e: 'f', biasing: { biases }};
-        const searchRequest = { e: 'f' };
+        const searchRequest = <any>{ e: 'f' };
```
jmbtutor: It might be preferable for us to write it like this instead just for style, like with `flux` a few lines down, but it doesn't really matter.
```js
const searchRequest: any = { e: 'f' };
```

PR #181, opened 2018-07-23
fix(SF-1009): Don't make a new object for refinements that don't change

Vic-Dev:


test/unit/core/reducers/data/navigations.ts
```diff
@@ -75,6 +89,13 @@ suite('navigations', ({ expect }) => {
       expect(reducer).to.eql(newState);
     });
 
+    it('should return original refinement for refinements that weren\'t selected on RESET_REFINEMENTS', () => {
+      const reducer = navigations(state, { type: Actions.RESET_REFINEMENTS, payload: true });
+
+      expect(reducer.byId.Format).to.not.eql(state.byId.Format);
+      expect(reducer.byId.Department).to.eql(state.byId.Department);
```
jmbtutor: Do you mean `eq` for both? I think you want to check the reference, not their contents.

Vic-Dev: good point. fixed

PR #182, opened 2018-07-24
fix(SF-1021): Apply real-time biasing immediately to autocomplete

jmbtutor:


src/core/requests.ts
```diff
@@ -1,4 +1,4 @@
-import { Request } from 'groupby-api';
+import { Request, SelectedRefinement } from 'groupby-api';
```
Vic-Dev: Remove unused type

PR #183, opened 2018-07-25
feat(SF-1026): Add option to always use the default collection on search

jmbtutor:
Set the configuration option `search.useDefaultCollection` to `true` to always search against the default collection, or `false` to search against the current selected collection. The default is `false`.

### DIFF
--- a/src/core/selectors.ts
+++ b/src/core/selectors.ts
@@ -47,6 +47,14 @@ namespace Selectors {
   };
 
   /**
+   * Returns the name of the default collection
+   */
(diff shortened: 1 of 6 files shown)


src/core/selectors.ts
```diff
@@ -46,6 +46,14 @@ namespace Selectors {
     return corrected || original;
   };
 
+  /**
+   * Returns the name of the default collection
+   */
+  export const defaultCollection = (state: Store.State) => {
```
Vic-Dev: Need to test this :)

PR #189, opened 2018-08-08
feat(SF-1015): Make storing the search response in the store configurable

Vic-Dev:


Vic-Dev:
We're not doing this right now because we're not sure how to split up the requests

PR #192, opened 2018-08-10
feat(SF-1049): recordCount should have the ability to show full recordCount

Vic-Dev:


src/core/selectors.ts
```diff
@@ -561,6 +561,14 @@ namespace Selectors {
             ? { navigationName: nav.field, type: 'Range', high, low }
             : { navigationName: nav.field, type: 'Value', value }), [])), []);
 
+  /**
+   * Helper function to display record count limited by the max records we can
+   * return from the database.
+   */
+  export const getLimitedCountDisplay = (count: number): string => {
+    const limitedCount = Search.extractRecordCount(count);
+    return count <= limitedCount ? `${count}` : `${limitedCount}+`;
```
jmbtutor: As much as I feel that a template literal with just a variable in it feels wrong, the alternative is `count + ''` or `count.toString()`, so I'm okay with it.

test/unit/core/selectors.ts
```diff
@@ -1052,4 +1052,38 @@ suite('selectors', ({ expect, stub }) => {
...
+    });
+  });
+
+  describe('getLimitedCountDisplay()', () => {
+    it('should return the given count as a string', () => {
+      expect(Selectors.getLimitedCountDisplay(10000)).to.eq('10000');
+      expect(Selectors.getLimitedCountDisplay(1)).to.eq('1');
+    });
+
+    it('should return the string 1000+', () => {
```
jmbtutor: `10000+` (missing a `0`)

PR #195, opened 2018-09-05
fix(SF-1053): Flatten batch actions

jmbtutor:


test/unit/core/store/middleware.ts
```diff
@@ -422,4 +419,29 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+      const actions = [{ a: 'b' }, { c: 'd' }];
+      const thunkEvaluator = spy(Middleware, 'thunkEvaluator');
+
+      Middleware.batchMiddleware(store)(next)(actions);
+
+      expect(thunkEvaluator).to.be.calledTwice;
+      expect(next).to.be.calledWith(actions[0]);
+      expect(next).to.be.calledWith(actions[1]);
+    });
+  });
```
Vic-Dev: Can you add a test for flattening arrays of mixed object and functions that return objects?
And can you also add a test for getting an action that is a function that returns an object?

PR #196, opened 2018-09-07
feat(SF-1069): Modify fetch actions to accept an options object with request property

Vic-Dev:


src/core/actions/payloads.ts
```diff
@@ -67,14 +106,16 @@ namespace Payload {
     }
   }
 
+  export interface NavigationId {
```
jmbtutor: Would this be better put under the `Navigation` namespace and be called `Navigation.Id`?

src/core/actions/creators.ts
```diff
@@ -20,125 +20,283 @@ namespace ActionCreators {
...
+   * @param  {string} navigationId - The navigationId for the navigation to fetch more refinements against.
+   * @return {Actions.FetchMoreRefinements} - Action with { navigationId }.
    */
-  export function fetchMoreRefinements(navigationId: string): Actions.FetchMoreRefinements {
-    return createAction(Actions.FETCH_MORE_REFINEMENTS, navigationId);
+  export function fetchMoreRefinements(navigationId: string): Actions.FetchMoreRefinements;
+  // tslint:disable-next-line typedef
+  export function fetchMoreRefinements(options) {
+    if (typeof options === 'string') {
+      return createAction(Actions.FETCH_MORE_REFINEMENTS, { navigationId: options });
```
jmbtutor: To reduce code duplication, perhaps it would be better to build an options object and `return createAction` once at the end using the object that we build? Nothing else changes other than that object.

src/core/actions/index.ts
```diff
@@ -85,35 +85,35 @@ namespace Actions {
 
   // fetch actions
   export const FETCH_MORE_REFINEMENTS = 'FETCH_MORE_REFINEMENTS';
-  export type FetchMoreRefinements = Action<typeof FETCH_MORE_REFINEMENTS, string>;
+  export type FetchMoreRefinements = Action<typeof FETCH_MORE_REFINEMENTS, Payload.Fetch.MoreRefinements>;
   export const FETCH_PRODUCTS = 'FETCH_PRODUCTS';
-  export type FetchProducts = Action<typeof FETCH_PRODUCTS>;
+  export type FetchProducts = Action<typeof FETCH_PRODUCTS, Payload.Fetch.Override>;
   export const FETCH_PRODUCTS_WITHOUT_HISTORY = 'FETCH_PRODUCTS_WITHOUT_HISTORY';
   export type FetchProductsWithoutHistory = Action<typeof FETCH_PRODUCTS_WITHOUT_HISTORY>;
```
Vic-Dev: Add `Payload.Fetch.Override`

src/core/actions/index.ts
```diff
@@ -85,35 +85,35 @@ namespace Actions {
...
   export const FETCH_PRODUCTS_WHEN_HYDRATED = 'FETCH_PRODUCTS_WHEN_HYDRATED';
   export type FetchProductsWhenHydrated = Action<typeof FETCH_PRODUCTS_WHEN_HYDRATED, Actions.FetchProducts>;
```
Vic-Dev: Fix type

src/core/actions/payloads.ts
```diff
@@ -2,6 +2,33 @@ import Configuration from '../configuration';
 import Store from '../store';
 
 namespace Payload {
+  export namespace Fetch {
+    export interface Override {
+      request?: any;
+    }
+
+    export interface MoreRefinements extends NavigationId, Override {}
+
```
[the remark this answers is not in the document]
Vic-Dev: sounds good

src/core/actions/payloads.ts
```diff
@@ -23,12 +54,19 @@ namespace Payload {
   }
 
   export namespace Collection {
-    export interface Count {
+    export interface Name {
       collection: string;
+    }
+
+    export interface Count extends Name {
```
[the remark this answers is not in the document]
Vic-Dev: I think we should make `Collection` an interface that has `count: number;` and extends `Name`

src/core/actions/payloads.ts
Vic-Dev: Makes sense

src/core/actions/index.ts
Vic-Dev: nvm

src/core/actions/payloads.ts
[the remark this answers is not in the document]
Vic-Dev: nvm

src/core/actions/creators.ts
```diff
@@ -20,125 +20,262 @@ namespace ActionCreators {
   // fetch action creators
   /**
    * Makes a request for more refinements for given navigation.
-   * @param  {string}                       navigationId - The navigationId for
-   * the navigation to fetch more refinements against.
-   * @return {Actions.FetchMoreRefinements}              - Action with navigationId.
+   * @param  {Actions.Payload.Fetch.MoreRefinements} options - An object with the navigationId for
+   * the navigation to fetch more refinements against and a request object for override.
+   * @return {Actions.FetchMoreRefinements} - Action with { navigationId, request }.
```
jmbtutor: This is parsed as Markdown, so `{ navigationId, request }` should be in backticks. Same goes for the others.

PR #199, opened 2018-09-18
fix(SF-1083): Do not impose query restrictions on searches

jmbtutor:
Checking to see if the search is empty or different is business logic that may or may not be desired. Lift these restrictions so that repeat queries and empty queries can be sent.

### DIFF
--- a/src/core/actions/creators.ts
+++ b/src/core/actions/creators.ts
@@ -317,10 +317,7 @@ namespace ActionCreators {
     return [
       ActionCreators.resetPage(),
       createAction(Actions.UPDATE_QUERY, query && query.trim(), {
-        payload: [
-          validators.isValidQuery,
-          validators.isDifferentQuery
-        ]
+        payload: validators.isValidQuery,
       })
     ];
   }
--- a/src/core/actions/validators.ts
+++ b/src/core/actions/validators.ts
@@ -18,11 +18,6 @@ export const isValidQuery: Validator<string> = {
   msg: 'search term is empty'
 };
 
-export const isDifferentQuery: Validator<string> = {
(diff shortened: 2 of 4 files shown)


src/core/actions/creators.ts
```diff
@@ -316,12 +316,7 @@ namespace ActionCreators {
   export function updateQuery(query: string): Actions.ResetPageAndUpdateQuery {
     return [
       ActionCreators.resetPage(),
-      createAction(Actions.UPDATE_QUERY, query && query.trim(), {
-        payload: [
-          validators.isValidQuery,
```
Vic-Dev: bring `isValidQuery` back, afaik we still want to check that

src/core/actions/creators.ts
```diff
@@ -316,12 +316,7 @@ namespace ActionCreators {
...
-          validators.isDifferentQuery
```
Vic-Dev: Find out if anything else uses `isDifferentQuery` and delete it + tests if it's not used anymore

src/core/actions/creators.ts
```diff
@@ -420,7 +415,7 @@ namespace ActionCreators {
         actions.push(ActionCreators.selectCollection(Selectors.defaultCollection(state)));
       }
       // tslint:disable-next-line max-line-length
-      actions.push(...ActionCreators.resetRefinements(true), ...ActionCreators.updateQuery(query || Selectors.query(state)));
+      actions.push(...ActionCreators.resetRefinements(true), ...ActionCreators.updateQuery(typeof query === 'string' ? query : Selectors.query(state)));
```
Vic-Dev: don't need to check if it's a string here if we bring `isValidQuery` back

PR #200, opened 2018-09-19
fix(SF-1109): Defer observer notifications until batch action has completed

jmbtutor:
Use `redux-batch-enhancer` to handle batch actions properly. This library defers notifying the observers until there are no batch actions in progress.


src/core/store/middleware.ts
```diff
@@ -184,19 +187,10 @@ export namespace Middleware {
...
-      ),
-      applyMiddleware(
-        batchMiddleware,
-        ...middleware
-      ),
-      applyMiddleware(
-        batchMiddleware,
-        Middleware.validator
-      ),
+      applyMiddleware(...normalizingMiddleware, saveStateAnalyzer, pastPurchaseProductAnalyzer),
```
Vic-Dev: Should attempt to reduce the amount of `normalizingMiddleware` we apply

jmbtutor: We should do this as part of the URL ticket since we'll be touching it anyway.

PR #202, opened 2018-09-24
Sf 1073

zoomchik:


zoomchik:
Linked to https://github.com/groupby/storefront-navigation/pull/57

Vic-Dev (changes requested):
You need use a validator to check whether the navigationId is `orable` (ie, has the field `or: true` on it) before allowing this behaviour. You can't "select all" for `andable` navigations.

Vic-Dev (changes requested):
Needs tests

jmbtutor:
Archiving repo. Closing.

PR #208, opened 2018-10-09
feat(SF-1120): Add isFetching flags in the store

Vic-Dev:


test/unit/core/reducers/is-fetching.ts
```diff
@@ -0,0 +1,57 @@
...
+    it(`it should set ${section} to true on ${type}`, () => {
+      const reducer = isFetching(state, <any>{ type });
+
+      expect(reducer).to.eql({ ...state, [section]: true });
+    });
+  }
+
+  function expectDoneFetching(type: string, section: string) {
+    it(`it should set ${section} to false on ${type}`, () => {
+      const reducer = isFetching({ ...state, [section]: true }, <any>{ type });
```
jmbtutor: This value is not a reducer, but a reduced value. Perhaps `reduced` or `newState` is a better name?

PR #211, opened 2018-10-15
feat(SF-1123): Clean up duplicate middleware passes

Vic-Dev:


test/unit/core/reducers/index.ts
```diff
@@ -100,18 +100,36 @@ suite('reducers', ({ expect, stub }) => {
...
+  describe('undo', () => {
+    const state = { a: 'b' };
+    const present = { c: 'd' };
+
+    beforeEach(() => {
+      stub(dataReducers, 'default').returns(state);
+    });
+
+    reducers.HISTORY_UPDATE_ACTIONS.forEach((action) => {
+      it('should add to the past object on HISTORY_UPDATE_ACTIONS', () => {
```
jmbtutor: Change this string to use backticks and replace `HISTORY_UPDATE_ACTIONS` with `${action}`. We have no way of knowing which test this is otherwise.

PR #217, opened 2018-10-25
feat(SF-1152): Refactor pastPurchase saga to behave more like search

Vic-Dev:


src/core/sagas/recommendations.ts
```diff
@@ -109,48 +111,61 @@ export namespace Tasks {
...
+        });
+        const results = yield effects.call(Requests.search, flux, request);
+        const navigations = PastPurchaseAdapter.pastPurchaseNavigations(
+          yield effects.select(Selectors.config),
+          SearchAdapter.combineNavigations(results)
+        );
+        yield effects.put(<any>flux.actions.receivePastPurchaseRefinements(navigations));
+      }
+    } catch (e) {
+      return effects.put(flux.actions.receivePastPurchaseRefinements(e));
```
Vic-Dev: Should I add this action to the undo error actions? @jmbtutor

jmbtutor: If it affects the state, then yes, which it looks like it does.

test/unit/core/sagas/recommendations.ts
```diff
@@ -375,50 +375,78 @@ suite('recommendations saga', ({ expect, spy, stub }) => {
...
-        // });
+      it('should return if there are no pastPurchaseSkus', () => {
+        const flux: any = {};
+        const action: any = {};
+        const config = {};
+
+        const task = Tasks.fetchPastPurchaseNavigations(flux, action);
+
+        expect(task.next().value).to.eql(effects.select(Selectors.pastPurchases));
+        expect(task.next([]).value).to.be.undefined;
```
jmbtutor: It's better to test that `.done` is `true` if you want to check that a generator has returned.

src/core/sagas/recommendations.ts
Vic-Dev: just kidding it's already there

